PR #1, opened 2021-07-24
Support connectivity for workloads

masayag:
For workloads that define container host ports, the worker will be
responsible for creating a dedicated netfilter chain and add rules for
incoming IPv4 and IPv6 traffic for the desired ports.
When workload is being removed, the chain will be dropped.
On workload udpdates, the chain of rules will be re-created.

### DIFF
--- a/internal/workload/workload.go
+++ b/internal/workload/workload.go
@@ -2,22 +2,47 @@ package workload
 
 import (
 	"fmt"
-	"git.sr.ht/~spc/go-log"
-	api2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/api"
-	podman2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/podman"
-	"github.com/jakub-dzon/k4e-operator/models"
 	"io/ioutil"
-	v1 "k8s.io/api/core/v1"
 	"os"
 	"path"
-	"sigs.k8s.io/yaml"
 	"strings"
 	"time"
+
+	"git.sr.ht/~spc/go-log"
+	api2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/api"
+	"github.com/jakub-dzon/k4e-device-worker/internal/workload/network"
+	podman2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/podman"
+	"github.com/jakub-dzon/k4e-operator/models"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
 )
+
+const nfTableName string = "edge"
 
 type WorkloadManager struct {
 	manifestsDir string
-	workloads    api2.WorkloadAPI
+	workloads    *workloadWrapper
+}
+
+// workloadWrapper manages the workload and its configuration on the device
+type workloadWrapper struct {
+	workloads *podman2.Podman
+	netfilter *network.Netfilter
+}
+
+func newWorkloadWrapper() (*workloadWrapper, error) {
+	newPodman, err := podman2.NewPodman()
+	if err != nil {
+		return nil, err
+	}
+	netfilter, err := network.NewNetfilter()
+	if err != nil {
+		return nil, err
+	}
+	return &workloadWrapper{
+		workloads: newPodman,
+		netfilter: netfilter,
+	}, nil
 }
 
 func NewWorkloadManager(configDir string) (*WorkloadManager, error) {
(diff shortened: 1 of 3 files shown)


internal/workload/workload.go
```diff
@@ -195,5 +225,83 @@ func (w *WorkloadManager) toPodYaml(workload *models.Workload) ([]byte, error) {
...
+	hostPorts, err := getHostPorts(workload)
+	if err != nil {
+		return err
+	}
+	if len(hostPorts) == 0 {
+		return nil
+	}
+	//TODO: should verify if chain doesn't exist
+	if err := ww.netfilter.AddChain(nftTableName, workload.Name); err != nil {
+		// should we return nil and only log the error ?
```
pkliczewski, line 269: I think that network configuration is part of the workload. We need to validate this in the operator before attempting to deploy and here we fail and should not continue with the deployment. I think it is good to log here as well. It will help to understand why workload deployment failed.

internal/workload/workload.go
```diff
@@ -195,5 +225,83 @@ func (w *WorkloadManager) toPodYaml(workload *models.Workload) ([]byte, error) {
...
+		return err
+	}
+
+	// for workloads, a port will be opened for the pod based on hostPort
+	for _, p := range hostPorts {
+		rule := fmt.Sprintf("tcp dport %d ct state new,established counter accept", p)
+		if err := ww.netfilter.AddRule(nftTableName, workload.Name, rule); err != nil {
+			// should we return nil and only log the error ?
```
pkliczewski, line 277: In my opinion we should behave in similar way as for `addChain`. Log and return an error.

internal/workload/workload.go
masayag: iinw, the pod spec validates the correctness of the ports range, but not sure if we can do anything more than that on the operator side (as we don't expose any network configuration, at least not at this stage).
I'll return the error from this context and log it from its caller.

pkliczewski: I think what we can check is port collisions (if already used by other workload). Is there anything else which can cause `AddChain` to return an error?

masayag: the port collisions will not be detected by nft. you can have the same rule in more than one chain. e.g.:
```
table inet edge {
        chain nginx-sample-x {
                type filter hook input priority filter; policy accept;
                tcp dport 10000 ct state established,new counter packets 0 bytes 0 accept
        }

        chain nginx-sample1 {
                type filter hook input priority filter; policy accept;
                tcp dport 9980 ct state established,new counter packets 0 bytes 0 accept
        }

        chain nginx-sample2 {
                type filter hook input priority filter; policy accept;
                tcp dport 10000 ct state established,new counter packets 0 bytes 0 accept
        }
}
```

ports collision will be detected by podman. also, there is a chance other service might occupy the port (perhaps not managed by k4e operator - if such scenario makes any sense).
I'd suggest to address failures when running pods by podman, and in case of a failure to do clear any resources resources related to the workload. On next attempt to run the workload, those resources will be recreated.

On the management level, ports collision can be validated on cluster side, when adding or updating edgedeployment resources of the same edge device, and deny attempts to run edgedeployment  if a port is already taken by other edgedeployment.

masayag: the port collisions will not be detected by nft. you can have the same rule in more than one chain. e.g.:
```
table inet edge {
        chain nginx-sample-x {
                type filter hook input priority filter; policy accept;
                tcp dport 10000 ct state established,new counter packets 0 bytes 0 accept
        }

        chain nginx-sample1 {
                type filter hook input priority filter; policy accept;
                tcp dport 9980 ct state established,new counter packets 0 bytes 0 accept
        }

        chain nginx-sample2 {
                type filter hook input priority filter; policy accept;
                tcp dport 10000 ct state established,new counter packets 0 bytes 0 accept
        }
}
```

ports collision will be detected by podman. also, there is a chance other service might occupy the port (perhaps not managed by k4e operator - if such scenario makes any sense).
I'd suggest to address failures when running pods by podman, and in case of a failure to do clear any resources resources related to the workload. On next attempt to run the workload, those resources will be recreated.

On the management level, ports collision can be validated on cluster side, when adding or updating edgedeployment resources of the same edge device, and deny attempts to run edgedeployment  if a port is already taken by other edgedeployment.

pkliczewski (approved):
/lgtm

PR #3, opened 2021-08-11
Mapping pod name from spec to Podman pod ID

jakub-dzon:
Fix for ECOPROJECT-234.

This PR introduces file-backed mapping between pod name defined in the pod specification and ID assigned to the Podman pod.

As part of this change set `workloadWrapper` has been moved to a separate file.

### DIFF
--- a/internal/workload/manager.go
+++ b/internal/workload/manager.go
@@ -10,39 +10,14 @@ import (
 
 	"git.sr.ht/~spc/go-log"
 	api2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/api"
-	"github.com/jakub-dzon/k4e-device-worker/internal/workload/network"
-	podman2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/podman"
 	"github.com/jakub-dzon/k4e-operator/models"
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 )
 
-const nfTableName string = "edge"
-
 type WorkloadManager struct {
 	manifestsDir string
 	workloads    *workloadWrapper
-}
-
-// workloadWrapper manages the workload and its configuration on the device
-type workloadWrapper struct {
-	workloads *podman2.Podman
-	netfilter *network.Netfilter
-}
-
-func newWorkloadWrapper() (*workloadWrapper, error) {
-	newPodman, err := podman2.NewPodman()
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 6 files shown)


internal/workload/wrapper.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (ww workloadWrapper) Run(workload *v1.Pod, manifestPath string) error {
+	if err := ww.applyNetworkConfiguration(workload); err != nil {
+		return err
+	}
+	podIds, err := ww.workloads.Run(manifestPath)
+	if err != nil {
+		return err
+	}
+	return ww.mappingRepository.Add(workload.Name, podIds[0])
```
pkliczewski, line 78: I wonder what would happen if workload was run successfully and we failed to add mapping. I think we need to make it transactional either both are successful or none.

internal/workload/manager.go
```diff
@@ -10,47 +10,22 @@ import (
...
-	}, nil
-}
-
 func NewWorkloadManager(configDir string) (*WorkloadManager, error) {
 	manifestsDir := path.Join(configDir, "manifests")
 	if err := os.MkdirAll(manifestsDir, 0755); err != nil {
 		return nil, fmt.Errorf("cannot create directory: %w", err)
 	}
-	wrapper, err := newWorkloadWrapper()
+	wrapper, err := newWorkloadWrapper(configDir)
```
masayag, line 28: Do you think we need to execute this one only [once](https://golang.org/src/sync/once.go) to avoid any attempt to open the files more than once during the execution of the worker ?

internal/workload/wrapper.go
jakub-dzon: I agree with Moti - the complexity of compensating for errors in the `Run` method will only grow and the workload watchdog is a good place to make sure the state of the device is correct.

internal/workload/manager.go
jakub-dzon: No, I don't think it's needed - the initialization of the components can't happen more than once during device-workers instance life.

internal/workload/wrapper.go
jakub-dzon: I will change the `ensureWorkloadsFromManifestsAreRunning` method appropriately.

jakub-dzon: Fixed. @masayag, @pkliczewski - please have a look.

masayag (approved):

pkliczewski (approved):

PR #4, opened 2021-08-17
Parametrize podman socket name

pkliczewski:
I wanted to run the agent on rhel8.2 and noticed that `podman.socket` is not there. I found it under different name:

```bash
ss -l -p | grep podman
u_str  LISTEN  0  128  /run/podman/io.podman 13964167  * 0  users:(("systemd",pid=1,fd=87))
```

This PR provides a way to customize it.

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>

### DIFF
--- a/cmd/device-worker/main.go
+++ b/cmd/device-worker/main.go
@@ -3,6 +3,11 @@ package main
 import (
 	"context"
 	"fmt"
+	"net"
+	"os"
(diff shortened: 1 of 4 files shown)


pkliczewski:
It seems like this version is too old to work with anyway

PR #6, opened 2021-09-01
Attach export volume

jakub-dzon:
This PR is a first step in implementing data transfer functionality on the device side. 

Code in this PR injects `/export` hostpath mount to each container in each pod specification.  Workloads are expected to put there  files that need to be transferred to the control plane storage.

### DIFF
--- a/cmd/device-worker/main.go
+++ b/cmd/device-worker/main.go
@@ -23,6 +23,10 @@ import (
 
 var yggdDispatchSocketAddr string
 
+const (
+	defaultDataDir = "/var/local/yggdrasil"
+)
+
 func main() {
 	logLevel, ok := os.LookupEnv("LOG_LEVEL")
 	if !ok {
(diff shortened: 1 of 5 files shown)


internal/workload/manager.go
```diff
@@ -223,5 +230,17 @@ func (w *WorkloadManager) toPod(workload *models.Workload) (*v1.Pod, error) {
 	}
 	pod.Kind = "Pod"
 	pod.Name = workload.Name
+	exportVolume := volumes.HostPathVolume(w.volumesDir, workload.Name)
```
pkliczewski, line 233: Do we guarantee `workload.Name` to be unique?

jakub-dzon: Yes, it has to be, as it is a name of a CR.

pkliczewski: The name of a CR is unique only in context of specific namespace. I wonder whether only name is enough.

jakub-dzon: Yes. EdgeDevice and EdgeDeployment have to be in the same namespace as per current design. It is not possible to deploy workloads from different namespaces.

cmd/device-worker/main.go
```diff
@@ -23,6 +23,10 @@ import (
 
 var yggdDispatchSocketAddr string
 
+const (
+	defaultDataDir = "/var/local/yggdrasil"
```
masayag, line 27: Is there a need to update PoC docs about this ?

masayag (approved):

cmd/device-worker/main.go
jakub-dzon: Good point. Yes.

pkliczewski (approved):

PR #7, opened 2021-09-07
Send data to s3

jakub-dzon:
This PR introduces sub-system for synchronizing workload data files with s3 storage specified by the control plane.


internal/exec/exec.go
```diff
@@ -0,0 +1,28 @@
+package exec
+
+import (
+	"bytes"
+	"git.sr.ht/~spc/go-log"
+	"os/exec"
+)
+
+func RunProcess(command string, args []string, env []string) error {
+	cmd := exec.Command(command, args...)
```
pkliczewski, line 10: I wonder whether something similar to https://github.com/juntaki/bucketsync would make more sense than running a command?

jakub-dzon: I've used https://github.com/seqsense/s3sync instead.

pkliczewski: I suggested this project because the idea is interesting and it would run in process. Managing external processes is a bit more tricky to do it well.

jakub-dzon: As far as I understand the project you suggested is for mounting s3 to local fs and we would have to reconcile files on our own. Both original in-OS `aws s3 sync` and the library I eventually used instead to that for us.

jakub-dzon: OK.. After some thinking I think I understand how it could be used as it is...

internal/datatransfer/monitor.go
```diff
@@ -0,0 +1,107 @@
...
+func (m *Monitor) syncPaths() {
+	workloads, err := m.workloads.ListWorkloads()
+	if err != nil {
+		log.Errorf("Can't get the list of workloads: %v", err)
+	}
+	if len(workloads) == 0 {
+		return
+	}
+	storage := m.config.GetDeviceConfiguration().Storage
+	if storage != nil && storage.S3 != nil {
```
masayag, line 65: perhaps an early return if storage is nil or storage.s3 is nil to avoid one nesting level ?

masayag (approved):

pkliczewski (approved):

PR #9, opened 2021-09-22
ECOPROJECT-237 Removing undeployed workloads

jakub-dzon:
This PR introduces code that makes sure that workloads that are removed from configuration are removed from the device.


pkliczewski (approved):
/lgtm

PR #11, opened 2021-09-23
[ECOPROJECT-235] Resolve race condition 

jakub-dzon:
Resolves race condition between `WorkloadManagerUpdate` and `WorkloadManager.ensureWorkloadsFromManifestsAreRunning`.


pkliczewski (approved):
/lgtm

PR #13, opened 2021-09-24
[ECOPROJECT-266] Configure testing framework

jakub-dzon:
This PR introduces usage of Ginkgo as a testing framework and provides first two test suites.

Use `make test` to execute all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,11 @@
 LIBEXECDIR ?= /usr/libexec
+
+test-tools:
+	go get -t ./...
+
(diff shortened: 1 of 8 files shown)


pkliczewski (commented):
/lgtm
Should we run it when PR is pushed?

jakub-dzon:
> Should we run it when PR is pushed?

Yes, I will set it up after this PR is merged.

pkliczewski (approved):

PR #15, opened 2021-10-01
Retry device registration until configuration is provided by the control plane

jakub-dzon:
This PR introduces registration retry logic in order to address ECOPROJECT-245. The registration requests will be sent to the control plane until device receives its configuration for the first time (as an indication of a successful registration). 

Registration operation is idempotent so it won't affect cluster-side state if request is sent multiple times.

### DIFF
--- a/internal/hardware/hardware.go
+++ b/internal/hardware/hardware.go
@@ -2,7 +2,6 @@ package hardware
 
 import (
 	"encoding/json"
-	"git.sr.ht/~spc/go-log"
 	"github.com/jakub-dzon/k4e-operator/models"
 	"github.com/openshift/assisted-installer-agent/src/inventory"
 )
@@ -13,7 +12,7 @@ type Hardware struct {
 func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
 	inv := inventory.ReadInventory()
 
-	// Instead of copying filed-by-filed marshal to JSON and unmarshal to other struct
+	// Instead of copying field-by-field marshal to JSON and unmarshal to other struct
 	inventoryJson, err := json.Marshal(inv)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


internal/hardware/hardware.go
```diff
@@ -23,6 +22,6 @@ func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	log.Infof("Hardware info: %s", inventoryJson)
```
pkliczewski, line 26: Is this log entry not needed anymore?

internal/registration/registration.go
```diff
@@ -3,29 +3,63 @@ package registration
...
 	return &Registration{
 		hardware:         hardware,
 		os:               os,
 		dispatcherClient: dispatcherClient,
+		config:           config,
+		RetryAfter:       retryAfter,
+	}
+}
+
+func (r *Registration) RegisterDevice() {
```
pkliczewski, line 38: How the agent know that there is no need to talk to the operator if we are not registered? ATM even if we are not registered there is a lot of 404 is the logs for the agent attempts.

masayag (approved):

internal/hardware/hardware.go
masayag: If remained, it is going to be printed multiple times to the log (on each retry). I don't think this information is useful on the agent side. And if reported properly, will be stored on the edge-device CR when succeeded to register.

jakub-dzon: @masayag is right - that was the reason for the removal. That information isn't helpful being in the device-worker's logs.

internal/registration/registration.go
jakub-dzon: It doesn't. The gateway - yggdrasil starts polling its communication channels right after start - it can't be controlled by the  workers. I will be making sure that the HTTP errors are not logged on the lowest possible level (error will also be encountered when the device is "away").

pkliczewski: OK, let's explore with fleet mgmt how we could do it. In our context it doesn't make sense to poll if we are not registered.

pkliczewski (approved):

PR #16, opened 2021-10-02
ECOPROJECT-180: Adding a support of disconnect a device

bardielle:
https://github.com/jakub-dzon/yggdrasil/pull/4


internal/configuration/configuration.go
```diff
@@ -121,3 +122,11 @@ func (m *Manager) GetConfigurationVersion() string {
 func (m *Manager) IsInitialConfig() bool {
 	return m.initialConfig.Load().(bool)
 }
+func (m *Manager) DeleteDeviceConfig() error {
+	log.Infof("Removing device config file: %s", m.deviceConfigFile)
+	err := os.Remove(m.deviceConfigFile)
+	if err != nil {
+		log.Error(err)
```
masayag, line 129: should the error be ignored ? the expectation of  https://github.com/jakub-dzon/k4e-device-worker/pull/16/files#diff-8937ebacce65547ca6fcb202618610cee53b3011c0c6fa918ad383c5081004e4R50 is to get an error.

internal/workload/manager.go
```diff
@@ -233,6 +233,50 @@ func (w *WorkloadManager) RegisterObserver(observer Observer) {
...
+	err := os.RemoveAll(w.manifestsDir)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *WorkloadManager) DeleteVolumeDir() error {
+	err := os.RemoveAll(w.volumesDir)
```
masayag, line 263: in terms of consistency, pls add log here as well

internal/workload/manager.go
```diff
@@ -233,6 +233,50 @@ func (w *WorkloadManager) RegisterObserver(observer Observer) {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (w *WorkloadManager) DeleteTable() error {
+	err := w.workloads.RemoveTable()
```
masayag, line 272: in terms of consistency, pls add log here as well

internal/server/deviceserver.go
```diff
@@ -37,3 +40,16 @@ func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error
 	// Respond to the start request that the work was accepted.
 	return &pb.Receipt{}, nil
 }
+
+// Disconnect implements the "Disconnect" method of the Worker gRPC service.
+func (s *deviceServer) Disconnect(ctx context.Context, in *pb.Empty) (*pb.DisconnectResponse, error) {
+	log.Infof("received worker disconnect request")
```
masayag, line 46: just log.Info() when no formatting is needed

masayag (changes requested):
Could include in the PR message links to the other PRs (on yggdrasil and on operator)

internal/configuration/configuration.go
bardielle: you'r right fixed it

internal/server/deviceserver.go
```diff
@@ -37,3 +40,16 @@ func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error
 	// Respond to the start request that the work was accepted.
 	return &pb.Receipt{}, nil
 }
+
+// Disconnect implements the "Disconnect" method of the Worker gRPC service.
+func (s *deviceServer) Disconnect(ctx context.Context, in *pb.Empty) (*pb.DisconnectResponse, error) {
+	log.Info("received worker disconnect request")
```
pkliczewski, line 46: Will we accept new workload requests at this point? I think once we receive disconnect we should not accept any new workloads.

bardielle: It means I need to add a lock for that section and also for the creating

pkliczewski:
/lgtm
@bardielle please rebase

internal/deregistration/deregistration.go
```diff
@@ -0,0 +1,50 @@
+package deregistration
+
+import (
+	"git.sr.ht/~spc/go-log"
+	"github.com/jakub-dzon/k4e-device-worker/internal/configuration"
+	"github.com/jakub-dzon/k4e-operator/models"
+	"github.com/jakub-dzon/k4e-device-worker/internal/workload"
+)
+
+type Deregistration struct {
```
jakub-dzon, line 10: IMO that should be part of `registration.Registration` struct - the methods should belong to it.

internal/deregistration/deregistration.go
```diff
@@ -0,0 +1,50 @@
...
+func NewDeregistration(workloadsManager *workload.WorkloadManager, configManager *configuration.Manager) (*Deregistration, error) {
+	deregstration := Deregistration{
+		workloads:                   workloadsManager,
+		config:                      configManager,
+	}
+	return &deregstration, nil
+}
+
+func (d *Deregistration) Update(configuration models.DeviceConfigurationMessage) error {
+	err := d.workloads.RemoveAllWorkloads()
```
jakub-dzon, line 24: IMO each of the components should be responsible for executing its own de-registration logic; this package should not know what needs to be done in `d.workloads` and `d.config`. 
It should just call `d.workloads.Deregister`, `d.config.Deregister` and these methods should hide package-private complexity.

We can go even further and extend our config Observer interface with `Remove` method that would be called when needed by configuration module.

That would give us following flow:
- `deviceServer.Disconnect` calls `registration.Registration.Deregister`;
- `registration.Registration.Deregister` calls `configuration.Manager.DeleteDeviceConfig`;
- `configuration.Manager.DeleteDeviceConfig` calls `configuration.Observer.Remove` for each registered observer
- each observer performs its own de-registration logic.

@bardielle, @masayag, @pkliczewski  - WDYT?

jakub-dzon (changes requested):

internal/deregistration/deregistration.go
```diff
@@ -0,0 +1,50 @@
...
+
+func NewDeregistration(workloadsManager *workload.WorkloadManager, configManager *configuration.Manager) (*Deregistration, error) {
+	deregstration := Deregistration{
+		workloads:                   workloadsManager,
+		config:                      configManager,
+	}
+	return &deregstration, nil
+}
+
+func (d *Deregistration) Update(configuration models.DeviceConfigurationMessage) error {
```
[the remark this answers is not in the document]
bardielle: Added it

internal/deregistration/deregistration.go
bardielle: Actually I thought it's better each components will implements its own deregistration step - I already pushed it as a new commit

About the new suggestion flow- I'm not sure Registration is the correct place because the register component dosen't create the things that Deregistration removes

masayag: I like the idea. If we'd like to have retries for deregistration logic, then it might have to be duplicated in each component.
Also, when deregistering, we need somehow to coordinate the order of the deregistration which might be reversed to the way we register the observers.

cmd/device-worker/main.go
```diff
@@ -97,14 +99,21 @@ func main() {
...
-	hbs := heartbeat2.NewHeartbeatService(c, configManager, wl, &hw, dataMonitor)
+	hbs := heartbeat2.NewHeartbeatService(dispatcherClient, configManager, wl, &hw, dataMonitor)
+
 	configManager.RegisterObserver(hbs)
 
+	dr, err := deregistration2.NewDeregistration(wl, configManager, hbs)
+	if err != nil {
+		log.Fatal(err)
+	}
+	configManager.RegisterObserver(dr)
```
jakub-dzon, line 110: `Deregistration` is registered as a configuration observer and Update method there is responsible for deregistering the device, which is not a desired behavior - the device will get de-registered as soon as it receives the first configuration update.

jakub-dzon (changes requested):

internal/deregistration/deregistration.go
jakub-dzon: I agree that it would be cumbersome to manage order of deregistration in observer pattern. 
@bardielle - nicely done! 
I still think of the `registration` package as something that manages all functionality related to device registration (state), which also includes de-registration and `deregistration.Deregistration.Update` should be renamed and moved to `registration.Registration`

internal/workload/manager.go
```diff
@@ -234,6 +234,82 @@ func (w *WorkloadManager) RegisterObserver(observer Observer) {
 	w.workloads.RegisterObserver(observer)
 }
 
+func (w *WorkloadManager) Deregister() error {
+	w.managementLock.Lock()
```
masayag, line 238: if we obtain the lock here, and from some reason in another thread, `workload.Update()` is waiting on the lock to be released, the call for `workload.Update()` will fail (either on attempt to read manifest folder, or at other step of its execution).
Eventually the error message will be propagated as warning to the logs.
But it can be avoided.
Also, we need to terminate `ensureWorkloadsFromManifestsAreRunning()` which will fail to read manifest folder

cmd/device-worker/main.go
```diff
@@ -97,14 +99,21 @@ func main() {
 	wl.RegisterObserver(dataMonitor)
 	dataMonitor.Start()
 
-	hbs := heartbeat2.NewHeartbeatService(c, configManager, wl, &hw, dataMonitor)
+	hbs := heartbeat2.NewHeartbeatService(dispatcherClient, configManager, wl, &hw, dataMonitor)
```
masayag, line 102: we need to terminate also dataMonitor that runs its `syncPaths()` on its own goroutine

internal/workload/manager.go
```diff
@@ -234,6 +234,82 @@ func (w *WorkloadManager) RegisterObserver(observer Observer) {
...
+	if err != nil {
+		log.Errorf("failed to delete manifests directory: %v", err)
+	}
+
+	err = w.deleteTable()
+	if err != nil {
+		log.Errorf("failed to delete table: %v", err)
+	}
+
+	err = w.deleteVolumeDir()
```
masayag, line 256: @jakub-dzon wouldn't we like to make sure datatransfer/monitor.syncPaths() is completed before this step to avoid data loss ?

masayag (changes requested):

internal/workload/manager.go
jakub-dzon: Good point.  We need to make at least one sync after all the workloads have been stopped.

internal/deregistration/deregistration.go
bardielle: One thing that I'm not sure about - the order of deregistration for each observer it improtant (cant remove the nftable or the volumes before deleting the workloads.)
How can I manage the dependencies?

jakub-dzon: I think overall structure is good as it is and I don't see value anymore in using observer pattern in deregistration (due to coordination problems that you and Moti pointed out). The only thing I would like to be different is where deregistration logic is implemented (`registration` package instead of `registration`).

internal/workload/manager.go
bardielle: why? 
If another thread `workload.Update()` will wait till the lock would be releases.
Then - the section of https://github.com/jakub-dzon/k4e-device-worker/blob/537d75ce04f05eddbbbd6859bfad0877b9170688/internal/workload/manager.go#L81 would not be run (there isn't any workloads) and also deployedWorkloadByName would be empty so the section  https://github.com/jakub-dzon/k4e-device-worker/blob/537d75ce04f05eddbbbd6859bfad0877b9170688/internal/workload/manager.go#L121
would not be run also....

go.mod
```diff
@@ -31,4 +31,7 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.21.0
 	sigs.k8s.io/cluster-api-provider-aws => github.com/openshift/cluster-api-provider-aws v0.2.1-0.20201022175424-d30c7a274820
 	sigs.k8s.io/cluster-api-provider-azure => github.com/openshift/cluster-api-provider-azure v0.1.0-alpha.3.0.20201016155852-4090a6970205
+
 )
+
+replace github.com/redhatinsights/yggdrasil => github.com/bardielle/yggdrasil v0.0.0-20211003122930-1782a292da22
```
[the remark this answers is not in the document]
bardielle: I will open a relevant PR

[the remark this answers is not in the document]
jakub-dzon: :+1: 
I'm waiting with reviewing/merging of the yggrdasil PR to keep compatibility (new yggdrasil won't work with old worker, but new worker will work with old yggdrasil)

internal/workload/manager.go
masayag: Do you refer to the finalizers that assures the disconnect is being called only if YggdrasilWorkloadFinalizer was removed ?
In that case - there is a point. That would be the responsibility of the operator not to send any deployments to the device if it has not YggdrasilWorkloadFinalizer finalizer.

internal/heartbeat/heartbeat.go
```diff
@@ -121,3 +121,11 @@ func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) error {
 	s.initTicker(periodSeconds)
 	return nil
 }
+
+func (s *Heartbeat) Deregister() error {
+	log.Info("Stopping heartbeat ticker")
```
masayag, line 126: Can you add to initTicker a message when the loop is exited that the heartbeat is stopped ?

internal/datatransfer/monitor.go
```diff
@@ -39,6 +39,14 @@ func (m *Monitor) Start() {
 	}()
 }
 
+func (m *Monitor) Deregister() error {
+	log.Info("Stopping monitor ticker")
+	if m.ticker != nil {
+		m.ticker.Stop()
+	}
+	return nil
+}
```
masayag, line 48: Can you add to the Start() function a message that the monitor is stopped when the for loop is exited ?

internal/workload/manager.go
bardielle: As I understood - after removing each workloads there is a call to WorkloadRemoved for each observer
So I think that no need to add "sync" call... 
@masayag @jakub-dzon what do you think?

bardielle: As I understood - after removing each workloads there is a call to WorkloadRemoved for each observer
So I think that no need to add "sync" call...
@masayag  @jakub-dzon what do you think?

jakub-dzon: Yes, there doesn't need to be an explicit call to sync data, but `WorkloadRemoved` method needs to be modified to execute that one last sync for the workload.

internal/workload/manager.go
bardielle: I added a boolean for that case. 
Please take a look

masayag:
LGTM
@jakub-dzon if you approve the PR, make sure to coordinate the merge on two repositories.

masayag (approved):

internal/datatransfer/monitor.go
```diff
@@ -49,6 +58,8 @@ func (m *Monitor) GetLastSuccessfulSyncTime(workloadName string) *time.Time {
 }
 
 func (m *Monitor) WorkloadRemoved(workloadName string) {
+	m.syncPaths()
```
jakub-dzon, line 61: It should be triggered for one specific workload, not all of them.

internal/registration/registration.go
```diff
@@ -87,3 +96,26 @@ func (r *Registration) registerDeviceOnce() error {
 	}
 	return nil
 }
+
+func (r *Registration) Deregister(configuration models.DeviceConfigurationMessage) error {
```
jakub-dzon, line 100: This method doesn't need a parameter (`configuration models.DeviceConfigurationMessage`).

internal/datatransfer/monitor.go
bardielle: Im not sure it's better that 
If we will add a map in the configuration.Manager we will need to update that map in each create/remove/update workload - it will be place that holds the data on workloads (this will be a duplicate struct)

masayag: According to the sequence of actions of wrapper.Remove(workload) - something will now work properly here.
I'll try to simplify it by removing unnecessary code:
```
func (ww workloadWrapper) Remove(workloadName string) error {
	ww.workloads.Remove(workload)
	ww.mappingRepository.Remove(workloadName)

	for _, observer := range ww.observers {
		observer.WorkloadRemoved(workloadName)
	}
}
```

meaning, the call for `m.syncPaths()` is being invoked only after the workload is removed.
however, `m.syncPaths()` starts with:
```
func (m *Monitor) syncPaths() {
	workloads, err := m.workloads.ListWorkloads()
	if len(workloads) == 0 {
		return
	}
```
At least for the last workload, this will quit without last sync for its data.

I'd try to simplify it and don't attempt to sync data as part of each remove.
After all, we'd might end up running `syncPaths()` here in parallel to the one being invoked by the ticker.

I can suggest an alternative approach so we won't have to mess to much with this code and make it too complex. See my other comment on manager.Deregister: https://github.com/jakub-dzon/k4e-device-worker/pull/16/files#diff-ab1f64e13b34d19039deb0762b992d04e2ece198e03ee84091e3df02a8d776b3R101

internal/registration/registration.go
```diff
@@ -87,3 +96,26 @@ func (r *Registration) registerDeviceOnce() error {
...
+	err := r.workloads.Deregister()
```
masayag, line 101: For the `syncPaths()` we can return here the list of the stopped workloads and provide them to `r.monitor.Deregister()`
which will shutdown the ticker and will call synchronously the `syncPaths()` with the given list of workloads that need to be synced.

We can even to monitor.StopSyncPaths() (new function) when this function starts to stop the ticker, then continue with the steps above.

In this scenario, assuming I'm not missing anything, a single syncPaths(workloads) is enough.

masayag (changes requested):

internal/datatransfer/monitor.go
```diff
@@ -49,10 +58,69 @@ func (m *Monitor) GetLastSuccessfulSyncTime(workloadName string) *time.Time {
...
+	storage := m.config.GetDeviceConfiguration().Storage
+	if storage == nil || storage.S3 == nil {
+		return
+	}
+
+	workloads, err := m.workloads.ListWorkloads()
+	if err != nil {
+		log.Errorf("Can't get the list of workloads: %v", err)
+	}
+	if len(workloads) == 0 {
```
jakub-dzon, line 78: This check will fail when `WorkloadRemoved` is called for the last workload and the date will not be synchronized for it.
`if len(dataPaths) == 0 {` check few lines is enough to finish execution of the method early and `len(workloads) == 0` check can be removed.

jakub-dzon (commented):
One last comment. The rest looks good:+1:

Please add tests for the added and changed code (in a separate PR).

jakub-dzon:
@bardielle - please rebase.

masayag (approved):

PR #23, opened 2021-10-05
Tiny improvements

eloycoto:
Two commits in this PR, not sure if it's the way that you used to work, but detailed information on each commit message:

1) Small change on Makefile to make it easier to run on fedora.
2) Small logging changes to add context to error messages.

Thanks!

### DIFF
--- a/cmd/device-worker/main.go
+++ b/cmd/device-worker/main.go
@@ -74,7 +74,7 @@ func main() {
 	// Listen on the provided socket address.
 	l, err := net.Listen("unix", r.GetAddress())
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalf("Cannot start listening on %s err: %v", r.GetAddress(), err)
 	}
 
 	// Register as a Worker service with gRPC and start accepting connections.
@@ -87,7 +87,7 @@ func main() {
 
 	wl, err := workload2.NewWorkloadManager(dataDir)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal("Cannot start Workload Manager, err: ", err)
 	}
 	configManager.RegisterObserver(wl)
 
@@ -112,6 +112,6 @@ func main() {
 	}
 
 	if err := s.Serve(l); err != nil {
-		log.Fatal(err)
+		log.Fatal("Cannot start worker server, err:", err)
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/device-worker/main.go
```diff
@@ -87,7 +87,7 @@ func main() {
 
 	wl, err := workload2.NewWorkloadManager(dataDir)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal("Cannot started Workload Manager, err: ", err)
```
jakub-dzon, line 90: "Cannot start ..."

cmd/device-worker/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 	// Listen on the provided socket address.
 	l, err := net.Listen("unix", r.GetAddress())
 	if err != nil {
-		log.Fatal(err)
+		log.Fatal("Cannot start listening on %s err: ", r.GetAddress(), err)
```
jakub-dzon, line 77: I think that should be `Fatalf`.

jakub-dzon (changes requested):
Thank you for the first contribution!

cmd/device-worker/main.go
masayag: I think there will be a mismatch between the number of arguments in the format string to the number of provided arguments.
Perhaps change to:
```
log.Fatalf("Cannot start listening on %s err: %v", r.GetAddress(), err)
```

cmd/device-worker/main.go
eloycoto: Fixed sorry!

cmd/device-worker/main.go
eloycoto: Fixed sorry!

cmd/device-worker/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 	// Listen on the provided socket address.
 	l, err := net.Listen("unix", r.GetAddress())
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalf("Cannot start listening on %s err: ", r.GetAddress(), err)
```
masayag, line 77: I think there will be a failure in printing the message due to a mismatch between the number of expected argument in the format message (1) to the actual number or provided arguments (2)

masayag (approved):

PR #24, opened 2021-10-05
ECOPROJECT-267: Refactor command flags

jordigilh:
This PR introduces 2 changes:
* (ECOPROJECT-267) Parametrize the workload watchdog time interval by exposing environment variable `WATCHDOG_INTERVAL_SECONDS` and flag  `--watchdog-interval`.

And as a side effect:
* Rework the `device-worker` flags with backwards compatibility: values can be defined with their existing environment variables or via command flags.

@jakub-dzon @masayag @eloycoto @pkliczewski please take a look and let me know what you think.


cmd/device-worker/main.go
```diff
@@ -115,3 +108,37 @@ func main() {
 		log.Fatal("Cannot start worker server, err:", err)
 	}
 }
+
+var (
```
masayag, line 112: Could you move the variables definition before the main ?
We don't have any convention for that (nor we don't use formatter), but it feels more convenient to see the variable definition before they are being used.

masayag (commented):
Before conducting a deeper review, I'd like to ask @jakub-dzon if he sees a need in supporting flags.
After all, yggd service is the only one who is going to invoke the device-worker and he does it in a certain way using env-variables.
I'm not familiar with use-case that requires flags.

jakub-dzon:
That setting should be controlled by the user in `EdgeDevice` CR and sent over as a part of expected configuration (and changed live in running device-worker) - like the Heartbeat setting.

jakub-dzon:
@masayag is right about the env vs. flag usage - only yggdrasil starts the device-worker and uses env variables to pass some simple settings to the workers (same for every worker); let's not change that contract to flags in general and for device-worker specific settings in particular:)

jordigilh:
> That setting should be controlled by the user in `EdgeDevice` CR and sent over as a part of expected configuration (and changed live in running device-worker) - like the Heartbeat setting.

What do you mean by "changed live in running device-worker?". Is it ok to use the environment variable to define the interval?

jakub-dzon:
@jordigilh - no. You need to extend https://github.com/jakub-dzon/k4e-operator/blob/47902a40f7a31b2ff882534bda6b528a0bff14b9/api/v1alpha1/edgedevice_types.go#L40 with interval setting and everything else to make it visible on the k4e-device-worker side (transport model in swagger, copying data from one model to the other) and then you need to apply new interval when the new configuration is received; see https://github.com/jakub-dzon/k4e-device-worker/blob/89f8f214e7f981ebf635cd7b1381da400a496858/internal/heartbeat/heartbeat.go#L116.

cmd/device-worker/main.go
jordigilh: I left it as it was.

masayag (approved):

PR #30, opened 2021-10-11
Workloads: Refactor and tests

eloycoto:
Workload Manager was not tested. Also, some features were not complete,
so this commits changes the following:

- `Observer.Update` interface now returns an array of errors instead of
  a simple error. Some workloads can work, others cannot.
- Added test for `workload.Manager` where almost all important cases are
  covered.
  
Fix ECOPROJECT-290

### DIFF
--- a/internal/configuration/configuration.go
+++ b/internal/configuration/configuration.go
@@ -2,14 +2,17 @@ package configuration
 
 import (
 	"encoding/json"
-	"git.sr.ht/~spc/go-log"
-	"github.com/jakub-dzon/k4e-operator/models"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"reflect"
 	"sync/atomic"
 	"time"
+
+	"git.sr.ht/~spc/go-log"
+	"github.com/hashicorp/go-multierror"
+	"github.com/jakub-dzon/k4e-operator/models"
 )
 
 var (
@@ -80,33 +83,39 @@ func (m *Manager) Update(message models.DeviceConfigurationMessage) error {
 	configurationEqual := reflect.DeepEqual(message.Configuration, m.deviceConfiguration.Configuration)
 	workloadsEqual := reflect.DeepEqual(message.Workloads, m.deviceConfiguration.Workloads)
 	log.Tracef("Initial config: [%v]; workloads equal: [%v]; configurationEqual: [%v]", m.IsInitialConfig(), workloadsEqual, configurationEqual)
+	var errors error
+
 	if m.IsInitialConfig() || !(configurationEqual && workloadsEqual) {
 		log.Tracef("Updating configuration: %v", message)
 		for _, observer := range m.observers {
 			err := observer.Update(message)
 			if err != nil {
-				return err
+				errors = multierror.Append(errors, fmt.Errorf("cannot update observer: %s", err))
+				return errors
 			}
 		}
 
 		// TODO: handle all the failure scenarios correctly; i.e. compensate all the changes that has already been introduces.
 		file, err := json.MarshalIndent(message, "", " ")
 		if err != nil {
-			return err
+			errors = multierror.Append(errors, fmt.Errorf("cannot unmarshal JSON: %s", err))
+			return errors
 		}
 		log.Tracef("Writing config to %s: %v", m.deviceConfigFile, file)
 		err = ioutil.WriteFile(m.deviceConfigFile, file, 0640)
 		if err != nil {
 			log.Error(err)
-			return err
+			errors = multierror.Append(errors, fmt.Errorf("cannot write device config file '%s': %s", m.deviceConfigFile, err))
+			return errors
 		}
 		m.deviceConfiguration = &message
 		m.initialConfig.Store(false)
 	} else {
 		log.Trace("Configuration didn't change")
+		return nil
 	}
 
-	return nil
+	return errors
 }
 
 func (m *Manager) GetDataTransferInterval() time.Duration {
--- a/internal/server/deviceserver.go
+++ b/internal/server/deviceserver.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+
 	"git.sr.ht/~spc/go-log"
 	configuration2 "github.com/jakub-dzon/k4e-device-worker/internal/configuration"
 	registration2 "github.com/jakub-dzon/k4e-device-worker/internal/registration"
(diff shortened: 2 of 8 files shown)


internal/configuration/configuration.go
```diff
@@ -75,37 +77,41 @@ func (m *Manager) GetWorkloads() models.WorkloadList {
...
+				errors = append(errors, fmt.Errorf("Cannot update observer, err: %s", err))
+				return errors
 			}
 		}
 
 		// TODO: handle all the failure scenarios correctly; i.e. compensate all the changes that has already been introduces.
 		file, err := json.MarshalIndent(message, "", " ")
 		if err != nil {
-			return err
+			errors = append(errors, fmt.Errorf("Cannot unmarshal Json, err: %s", err))
```
jordigilh, line 98: s/Json/JSON
You can find more here: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
Also, no need to mention ", err:", just a colon will do :smile: 
```
fmt.Errorf("cannot unmarshal JSON: %s", err)
```
WDYT?

internal/heartbeat/heartbeat.go
```diff
@@ -97,14 +97,14 @@ func (s *Heartbeat) HasStarted() bool {
...
-func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) error {
+func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) []error {
 	periodSeconds := s.getInterval(*config.Configuration)
 	log.Infof("Reconfiguring ticker with interval: %v", periodSeconds)
 	if s.ticker != nil {
 		s.ticker.Stop()
 	}
 	s.initTicker(periodSeconds)
-	return nil
+	return []error{}
```
jordigilh, line 107: you could create your own type for this... just a thought:

```
type errors []error
```

internal/workload/manager.go
```diff
@@ -77,49 +77,58 @@ func (w *WorkloadManager) GetExportedHostPath(workloadName string) string {
...
 	for _, workload := range configuration.Workloads {
 		log.Tracef("Deploying workload: %s", workload.Name)
 		configuredWorkloadNameSet[workload.Name] = struct{}{}
-		// TODO: change error handling from fail fast to best effort (deploy as many workloads as possible)
+
 		pod, err := w.toPod(workload)
 		if err != nil {
-			return err
+			errors = append(errors, fmt.Errorf(
+				"Cannot convert workload '%s' to pod, err: %v", workload.Name, err))
```
jordigilh, line 92: same comment as earlier about capitalization and discarding the `, err` substring

internal/server/deviceserver.go
```diff
@@ -28,9 +30,9 @@ func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error
 		if err != nil {
 			log.Warnf("Cannot unmarshal message: %v", err)
 		}
-		err = s.configManager.Update(deviceConfigurationMessage)
-		if err != nil {
-			log.Warnf("Failed to process message: %v", err)
+		errors := s.configManager.Update(deviceConfigurationMessage)
+		if len(errors) > 0 {
+			log.Warnf("Failed to process message: %v", fmt.Sprintf("%s", errors))
```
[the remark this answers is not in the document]
pkliczewski: Why not to use only `%v` as describe [here](https://pkg.go.dev/fmt)? It seems to have the same output with and without `fmt.Sprintf`

internal/configuration/configuration.go
```diff
@@ -80,33 +83,39 @@ func (m *Manager) Update(message models.DeviceConfigurationMessage) error {
...
+	var errors error
+
 	if m.IsInitialConfig() || !(configurationEqual && workloadsEqual) {
 		log.Tracef("Updating configuration: %v", message)
 		for _, observer := range m.observers {
 			err := observer.Update(message)
 			if err != nil {
-				return err
+				errors = multierror.Append(errors, fmt.Errorf("cannot update observer: %s", err))
+				return errors
```
jakub-dzon, line 94: This will stop the execution on first error; probably should be moved outside of the for-loop

internal/configuration/configuration.go
```diff
@@ -80,33 +83,39 @@ func (m *Manager) Update(message models.DeviceConfigurationMessage) error {
...
+			errors = multierror.Append(errors, fmt.Errorf("cannot unmarshal JSON: %s", err))
+			return errors
 		}
 		log.Tracef("Writing config to %s: %v", m.deviceConfigFile, file)
 		err = ioutil.WriteFile(m.deviceConfigFile, file, 0640)
 		if err != nil {
 			log.Error(err)
-			return err
+			errors = multierror.Append(errors, fmt.Errorf("cannot write device config file '%s': %s", m.deviceConfigFile, err))
+			return errors
```
jakub-dzon, line 109: Same here; probably no return at this stage.

internal/workload/manager.go
```diff
@@ -135,12 +146,14 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
 			err := os.Remove(manifestPath)
 			if err != nil {
 				if !os.IsNotExist(err) {
-					return err
+					errors = multierror.Append(errors, fmt.Errorf("cannot remove existing manifest workload: %s", err))
+					return errors
```
jakub-dzon, line 150: Let's try to remove all manifest files and only then return errors if any.

internal/workload/manager.go
```diff
@@ -135,12 +146,14 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
...
 				}
 			}
 
 			if err := w.workloads.Remove(name); err != nil {
-				return err
+				errors = multierror.Append(errors, fmt.Errorf("cannot remove stale workload name='%s': %s", name, err))
+				return errors
```
jakub-dzon, line 156: Same here - let's try to do as much as possible.

internal/workload/manager_test.go
```diff
@@ -0,0 +1,266 @@
...
+			// Then
+			Expect(err).To(HaveOccurred())
+			Expect(wkManager).To(BeNil())
+		})
+	})
+
+	Context("Update", func() {
+		It("works as expected", func() {
+
+			workloads := []*models.Workload{}
```
jakub-dzon, line 67: That's also `// given`

jakub-dzon (changes requested):

internal/configuration/configuration.go
eloycoto: Discussed offline, I'll address this on a new PR regarding configuration tests

PR #34, opened 2021-10-12
Removing workload-mapping.json as part of deregistration

bardielle:


internal/workload/mapping/mapping.go
```diff
@@ -67,6 +69,20 @@ func (m *MappingRepository) Remove(name string) error {
...
 }
 
+func (m *MappingRepository) RemoveMappingFile() error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	log.Info("Deleting workload-mapping.json file")
+	err := os.RemoveAll(m.mappingFilePath)
+	if err != nil {
+		log.Errorf("failed to delete manifests directory: %v", err)
```
jakub-dzon, line 79: Please fix the log message.

internal/workload/mapping/mapping.go
```diff
@@ -67,6 +69,20 @@ func (m *MappingRepository) Remove(name string) error {
 	return m.persist()
 }
 
+func (m *MappingRepository) RemoveMappingFile() error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	log.Info("Deleting workload-mapping.json file")
```
masayag, line 76: could you replace hardcoded file name with the value of `m.mappingFilePath` so any future changes to that value will be reflected in an updated log message ?

eloycoto:
Just a note, it would be great if you add a test here, workloads/mapping is already covered ;-)

PR #36, opened 2021-10-13
Tests: internal/configuration unit test

eloycoto:
Added test for internal/configuration package

Fix: https://issues.redhat.com/browse/ECOPROJECT-292


internal/configuration/configuration_test.go
```diff
@@ -0,0 +1,354 @@
...
+		}
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
+		_ = os.Remove(datadir)
+	})
+
+	Context("NewConfigurationManager", func() {
+		It("Raise an error if device-config file is not valid json", func() {
```
jakub-dzon, line 75: The name suggests there should be some error raised, but the test does not check for any error. 
WDYT about something similar to `should use configuration when device-config file is not valid json`.

PR #37, opened 2021-10-14
Test: internal/registration test

eloycoto:
These changes tests Register/Deregister actions on the device worker


internal/registration/registration_test.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		It("Work as expected", func() {
+
+			// given
+			reg := registration.NewRegistration(
+				hw, os, dispatcherMock, configManager, hb, wkManager, monitor)
+
+			// then
+			dispatcherMock.EXPECT().Send(gomock.Any(), RegistrationMatcher()).Times(1)
```
pkliczewski, line 85: What do you think about validating data being send?

internal/registration/registration_test.go
```diff
@@ -0,0 +1,172 @@
...
+			reg.RetryAfter = 1
+
+			// then
+			dispatcherMock.EXPECT().Send(gomock.Any(), gomock.Any()).Return(
+				nil, fmt.Errorf("failed")).Times(1)
+			dispatcherMock.EXPECT().Send(gomock.Any(), RegistrationMatcher()).Times(1)
+
+			//  when
+			reg.RegisterDevice()
+			time.Sleep(2 * time.Second)
```
pkliczewski, line 105: Why do we need to wait? It always causes flaky behavior in CI. Maybe we could add some logic on wait_on_register?

internal/registration/registration_test.go
```diff
@@ -0,0 +1,172 @@
...
+			// when
+			err := reg.Deregister()
+
+			// then
+			Expect(err).To(HaveOccurred())
+
+		})
+
+	})
+
```
pkliczewski, line 156: What do you think about adding test for register -> deregister -> register again?

internal/workload/manager.go
```diff
@@ -276,38 +276,45 @@ func (w *WorkloadManager) Deregister() error {
...
 	err := w.removeAllWorkloads()
 	if err != nil {
+		errors = multierror.Append(errors, fmt.Errorf("failed to remove workloads: %v", err))
 		log.Errorf("failed to remove workloads: %v", err)
 	}
 
 	err = w.deleteManifestsDir()
 	if err != nil {
 		log.Errorf("failed to delete manifests directory: %v", err)
+		log.Errorf("failed to delete manifests directory: %v", err)
```
pkliczewski, line 289: shouldn't we have this instead?

```go
errors = multierror.Append(errors, fmt.Errorf("failed to delete manifests directory: %v", err))
```

internal/registration/registration_test.go
```diff
@@ -0,0 +1,172 @@
...
+})
+
```
bardielle, line 158: What do you think about adding a test of deregister without having device-config file in the device directory? it should pass 

Also re-Deregister (can happen when sending simultaneity delete edge-device)

internal/registration/registration_test.go
pkliczewski: I understand we are testing this [line](https://github.com/jakub-dzon/k4e-device-worker/blob/main/internal/registration/registration.go#L94). My question was whether it makes sense to verify data attribute.

internal/registration/registration_test.go
eloycoto: Ok, make sense what you're suggesting, I'll implement that

masayag: waiting for an event/action can be done by using https://onsi.github.io/gomega/#making-asynchronous-assertions
not sure if that will fit here, since the condition to wait for isn't public.

jakub-dzon (approved):

PR #38, opened 2021-10-18
Internal/datatransfer: added tests

eloycoto:
Adding tests for internal/datatrasnfer package, with the following
changes:

- Added some Public functions to monitor struct.
- Sync separation with S3, so now it's an interface.
- Make simple the sync parts.

Fix https://issues.redhat.com/browse/ECOPROJECT-291


internal/datatransfer/monitor.go
```diff
@@ -17,17 +20,23 @@ type Monitor struct {
...
 	monitor := Monitor{
 		workloads:                   workloadsManager,
 		config:                      configManager,
 		ticker:                      ticker,
 		lastSuccessfulSyncTimes:     make(map[string]time.Time),
 		lastSuccessfulSyncTimesLock: sync.RWMutex{},
 	}
+	err := monitor.InitStorage()
+	if err != nil {
+		log.Error("Cannot init the storage")
```
pkliczewski, line 38: In this case user would get correct `Monitor` struct. Should we really logging error here if so why not to return it in case of issue with storage initialization or maybe update it to warn or trace?

internal/datatransfer/monitor_test.go
```diff
@@ -0,0 +1,270 @@
...
+			// when
+			err := monitor.ForceSync()
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("one sync failed", func() {
+			// given
+
```
pkliczewski, line 138: no need to have empty line here

internal/datatransfer/s3/sync.go
```diff
@@ -20,40 +21,58 @@ var (
...
 	if err != nil {
 		log.Errorf("Can't decode AWS CA Bundle: %v", err)
+		return nil, err
 	}
 
-	endpoint := fmt.Sprintf("https://%s:%d", s3Config.BucketHost, s3Config.BucketPort)
+	sync.endpoint = fmt.Sprintf("https://%s:%d", s3Config.BucketHost, s3Config.BucketPort)
+	sync.bucketName = s3Config.BucketName
+	return &sync, nil
+
```
pkliczewski, line 57: no need for extra line here

internal/datatransfer/monitor.go
```diff
@@ -17,17 +20,24 @@ type Monitor struct {
...
 	ticker := time.NewTicker(configManager.GetDataTransferInterval())
+
 	monitor := Monitor{
 		workloads:                   workloadsManager,
 		config:                      configManager,
 		ticker:                      ticker,
 		lastSuccessfulSyncTimes:     make(map[string]time.Time),
 		lastSuccessfulSyncTimesLock: sync.RWMutex{},
 	}
+	err := monitor.InitStorage()
```
jakub-dzon, line 36: Having the storage initialized once will cause problems when the s3 configuration changes; the Monitor will continue sending files to the old bucket with old credentials while the configuration might specify different ones.
That was the reason, why the `s3` client was created in `syncPaths` function for each synchronization cycle.

The only way to refresh `fsSync` is to restart the device worker, which is not desired.

internal/datatransfer/monitor.go
```diff
@@ -112,52 +121,98 @@ func (m *Monitor) getPathsOfWorkload(workloadName string) []*models.DataPath {
...
+	storage := m.config.GetDeviceConfiguration().Storage
+	if storage == nil {
+		return false
+	}
+
+	return storage.S3 != nil
+}
+
+func (m *Monitor) InitStorage() error {
+	storage := m.config.GetDeviceConfiguration().Storage
```
jakub-dzon, line 138: When the device worker is started for the first time and `Monitor` is created, `m.config.GetDeviceConfiguration().Storage` will most likely be nil and the `fsSync` will never be configured correctly. At least until device-worker is restarted.

internal/datatransfer/monitor.go
```diff
@@ -112,56 +123,125 @@ func (m *Monitor) getPathsOfWorkload(workloadName string) []*models.DataPath {
...
 	}
-	storage := m.config.GetDeviceConfiguration().Storage
-	if storage != nil && storage.S3 != nil {
-		workloadToDataPaths := make(map[string][]*models.DataPath)
-		for _, wd := range m.config.GetWorkloads() {
-			if wd.Data != nil && len(wd.Data.Paths) > 0 {
-				workloadToDataPaths[wd.Name] = wd.Data.Paths
-			}
+
+	if ok := m.HasStorageDefined(); !ok {
```
jakub-dzon, line 171: `if !m.HasStorageDefined()`

internal/datatransfer/monitor.go
```diff
@@ -76,7 +81,13 @@ func (m *Monitor) syncPathsWorkload(workloadName string) {
...
 	}
 
-	s3sync, err := s3.NewSync(*storage.S3)
+	syncWrapper, err := m.getFsSync()
+	if err != nil {
+		log.Errorf("Error while getting s3 synchronizer: %v", err)
+		return
+	}
+
+	err = syncWrapper.Connect()
```
jakub-dzon, line 90: What will happen if `Connect` is called here while `SyncPath` is being executed in `syncPaths`? Will there be any problem?

internal/datatransfer/monitor_test.go
```diff
@@ -0,0 +1,351 @@
...
+			err := monitor.ForceSync()
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+	})
+
+	Context("WorkloadRemoved", func() {
+		It("work as expected", func() {
```
jakub-dzon, line 241: Please make the name of the test more descriptive.

internal/datatransfer/monitor.go
eloycoto: No, I do not think so, the pointer is copy, so shouldn't be a problem at all. Or I cannot see it.

jakub-dzon: The pointer is a copy of pointer stored in the struct, but it points at the same value so effectively `Connect` is called against the same object, which changes internal state of `syncWrapper`. But is that dangerous in any way?

eloycoto: Added a if on the s3/sync.go

https://github.com/jakub-dzon/k4e-device-worker/pull/38/files#diff-5a3f35566e47eb15cffc1c02a0b18f6bbbf48883663884e699ed09ca12942ca7R60

jakub-dzon (approved):

pkliczewski (approved):

internal/datatransfer/monitor.go
```diff
@@ -112,56 +123,125 @@ func (m *Monitor) getPathsOfWorkload(workloadName string) []*models.DataPath {
...
+		return false
+	}
+
+	return storage.S3 != nil
+}
+
+func (m *Monitor) SetStorage(fs FileSync) {
+	m.syncMutex.Lock()
+	m.fsSync = fs
+	m.syncMutex.Unlock()
```
masayag, line 142: nit: i like the pattern of `defer m.syncMutex.Unlock()`  right after we acquire the lock.
algought this is a short and clear function, but it will be consistent with the rest of the code and will reduce changes of surprises in the future if this function will be changed

internal/datatransfer/monitor.go
```diff
@@ -112,56 +123,125 @@ func (m *Monitor) getPathsOfWorkload(workloadName string) []*models.DataPath {
...
-func (m *Monitor) syncPaths() {
 	workloads, err := m.workloads.ListWorkloads()
 	if err != nil {
 		log.Errorf("Can't get the list of workloads: %v", err)
+		return err
 	}
+
 	if len(workloads) == 0 {
-		return
+		log.Trace("No worloads to return")
```
masayag, line 167: s/worloads/workloads

internal/datatransfer/monitor_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+			monitor := datatransfer.NewMonitor(wkManager, configManager)
+			monitor.SetStorage(fssync)
+
+			wkwMock.EXPECT().List().Return([]api.WorkloadInfo{{
+				Id:     "test",
+				Name:   "test",
+				Status: "Running",
+			}}, nil).AnyTimes()
+			// when
```
masayag, line 129: pls add empty line before

internal/datatransfer/monitor_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+			// then
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("Failed with invalid s3 configuration", func() {
+			// given
+			monitor := datatransfer.NewMonitor(wkManager, configManager)
+			cfg.Configuration.Storage.S3.AwsAccessKeyID = "invalidbase64"
+			// when
```
masayag, line 343: pls add empty line before

masayag (changes requested):

masayag (approved):

PR #41, opened 2021-11-02
CI: Push rpm to fedora copr

eloycoto:
RPM was not created correctly, because it ships a binary when this is
    not allowed. With this change, the RPM is created from source, and
    src.rpm is created.

    Also, a new workflow is added, if any new commit on `main` branch, it'll
    build the src pacakge, and push the change to Fedora copr[0], where ARM
    and x86 builds are triggered.

    https://copr.fedorainfracloud.org/coprs/eloyocoto/k4e-test/packages/

@TODO: I need to document how to use this repo, but want to work on r4e image first to get full context. 

### DIFF
--- a/k4e-agent.spec
+++ b/k4e-agent.spec
@@ -4,7 +4,7 @@ Name:       k4e-agent
 Version:    1.0
 Release:    1%{?dist}
 Summary:    Agent application for the K4E (Kubernetes for Edge) solution
-
+ExclusiveArch: %{go_arches}
 Group:      K4E
 License:    GPL
 Source0:    https://github.com/jakub-dzon/%{name}-%{version}.tar.gz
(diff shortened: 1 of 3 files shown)


pkliczewski (approved):

k4e-agent.spec
```diff
@@ -25,9 +25,27 @@ The K4E agent communicates with the K4E control plane. It reports the status of
...
 systemctl enable --now nftables.service
 
+%prep 
+tar fx %{SOURCE0}
+
+%build
+cd k4e-agent-%{VERSION}
+export CGO_ENABLED=0 
+export GOFLAGS="-mod=vendor -tags=containers_image_openpgp"
+go build -o ./bin/device-worker ./cmd/device-worker
```
jakub-dzon, line 35: In case of ARM 64 `GOARCH=arm64` is not set; will it be built correctly if the only indication of the architecture is `ExclusiveArch`?

eloycoto: Mainly, on copr will detect that it's ARM host, and it'll run in there. ;-)

jakub-dzon (approved):

masayag (approved):

PR #45, opened 2021-11-03
Add missing call to rpm-tarball from rpm-src

masayag:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ rpm-tarball:
 	 (git archive --prefix k4e-agent-$(VERSION)/ HEAD ) \
 	    | gzip > k4e-agent-$(VERSION).tar.gz
 
-rpm-src:
+rpm-src: rpm-tarball
 	cp k4e-agent-$(VERSION).tar.gz $(HOME)/rpmbuild/SOURCES/
 	rpmbuild -bs \
 		-D "VERSION $(VERSION)" \


jakub-dzon (approved):

pkliczewski (approved):

eloycoto:
No needed at all, it's part of the action, but LGTM. 

I do not include just in case you want to test a bit faster.

eloycoto (approved):

masayag:
> No needed at all, it's part of the action, but LGTM.
> 
> I do not include just in case you want to test a bit faster.

It fails without it:

```
[root@kvm-06-guest34 k4e-device-worker]# make rpm-src
which: no ginkgo in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
which: no gover in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
cp k4e-agent-1.0.tar.gz /root/rpmbuild/SOURCES/
cp: cannot stat 'k4e-agent-1.0.tar.gz': No such file or directory
make: *** [Makefile:65: rpm-src] Error 1
```

eloycoto:
This is why I added in the action: 

https://github.com/jakub-dzon/k4e-device-worker/blob/main/.github/workflows/rpm.yaml#L31

When building/testing, I wanted to make it faster if the tarball was already created. 
As I said, it's ok ;-)

PR #47, opened 2021-11-09
Send event data about pod start failure

machacekondra:
This PR add possibility to collect information about failures during pod management actions. Currently we store this data in-memory so it won't persist node crash. So event data will be lost in case it was not sent and node crashes. We may introduce some file based queue if needed.

Currently by default start actions is executed every 15s and hb data are send ever 60s, so we try to start pod 4 times in minute. And we also send such event data to k8s. We may want to introduce increase of timeout in case pod fail to start. 15s, 30s, 60s, 120s, etc. same k8s does. But it may not be needed in case we use systemd for this.


machacekondra:
@masayag @pkliczewski @jakub-dzon @eloycoto PTAL

internal/workload/manager.go
```diff
@@ -238,6 +268,11 @@ func (w *WorkloadManager) ensureWorkloadsFromManifestsAreRunning() error {
 				// Workload is not running - start
 				err = w.workloads.Start(&podWithPath.pod)
 				if err != nil {
+					w.eventsQueue = append(w.eventsQueue, &models.EventInfo{
```
jakub-dzon, line 271: Are you planning to add events related to pod `run` failures as well in a separate PR?

internal/workload/manager.go
```diff
@@ -238,6 +268,11 @@ func (w *WorkloadManager) ensureWorkloadsFromManifestsAreRunning() error {
...
+						Message: err.Error(),
+						Reason:  "Started",
```
jakub-dzon, line 273: Shouldn't it be `NotStarted`? IMO `Started` would describe event of successful start of a pod.

machacekondra: Right. Actually k8s uses `Failed`, so I think I will reuse it.

internal/workload/manager.go
machacekondra: Sure I can.

pkliczewski (commented):
/lgtm

internal/workload/manager.go
pkliczewski: We can do it in separate PR

jakub-dzon (approved):

PR #49, opened 2021-11-10
Passing DEVICE_ID env variable to each container

jakub-dzon:
This PR exposes `DEVICE_ID` variable equal to ID under which the device is registered with the control plane to each container in every Pod.

### DIFF
--- a/internal/workload/manager_test.go
+++ b/internal/workload/manager_test.go
@@ -3,7 +3,9 @@ package workload_test
 import (
 	"fmt"
 	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
 	"os"
+	"sigs.k8s.io/yaml"
 	"time"
 
 	gomock "github.com/golang/mock/gomock"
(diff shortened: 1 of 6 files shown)


internal/workload/manager_test.go
```diff
@@ -92,6 +101,23 @@ var _ = Describe("Manager", func() {
 
 			// then
 			Expect(err).NotTo(HaveOccurred())
+			for i := 0; i < 10; i++ {
+				wkName := fmt.Sprintf("test%d", i)
+				pod := getPodFor(datadir, wkName)
+				Expect(pod.Name).To(BeEquivalentTo(wkName))
+
+				Expect(pod.Spec.Containers).To(HaveLen(1))
+				Expect(pod.Spec.Containers[0].Env).To(HaveLen(1))
```
eloycoto, line 110: Just a note, on loops in case of any fail, will be hard to know what pod failed, maybe adding some kind of message on the assert: 
```
Expect(pod.Spec.Containers[0].Env[0]).To(BeEquivalentTo(v1.EnvVar{Name: "DEVICE_ID", Value: deviceId}), fmt.Sprintf("Failing on pod %s", wkName))
```

jakub-dzon: Done

pkliczewski (commented):
/lgtm

eloycoto (approved):

masayag (approved):

PR #51, opened 2021-11-11
Datatransfer: Monitor force config load when device start

eloycoto:
When device start, the datatransfer/monitor didn't get the update at
all, because Update() was not called.

With this change, on start, we force the start of config load.

Fix ECOPROJECT-376

### DIFF
--- a/internal/datatransfer/monitor.go
+++ b/internal/datatransfer/monitor.go
@@ -38,6 +38,11 @@ func NewMonitor(workloadsManager *workload.WorkloadManager, configManager *confi
 }
 
 func (m *Monitor) Start() {
+	// Made this to force the config load in case there is no storage defined yet
+	// via Update() method.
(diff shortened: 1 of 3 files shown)


pkliczewski (approved):

internal/datatransfer/monitor.go
```diff
@@ -228,6 +228,13 @@ func (m *Monitor) storeLastUpdateTime(workloadName string) {
 	m.lastSuccessfulSyncTimes[workloadName] = time.Now()
 }
 
+func (m *Monitor) ForceUpdate() error {
```
jakub-dzon, line 231: WDYT about making it part of `Start`? After all it's supposed when the component starts.

masayag (approved):

jakub-dzon (approved):

PR #53, opened 2021-11-29
Custom authfiles

jakub-dzon:
This PR adds support for custom custom docker auth files sent by the control plane optionally with workloads.

This change allows users to schedule workloads that use images stored in private image repositories.


internal/workload/manager.go
```diff
@@ -195,6 +211,19 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
...
 }
 
+func (w *WorkloadManager) manageAuthFile(authFilePath, authFile string) (string, error) {
+	if authFile == "" {
+		if err := deleteFile(authFilePath); err != nil {
+			return "", fmt.Errorf("cannot remove auth file %s: %s", authFilePath, err)
+		}
+		return "", nil
+	}
+	if err := w.storeFile(authFilePath, []byte(authFile)); err != nil {
```
pkliczewski, line 221: For stored files how do we handle secret updates?

jakub-dzon: The secret gets overwritten if change is detected here: https://github.com/jakub-dzon/k4e-device-worker/pull/53/files#diff-8be2736004b01f09b6ffef40a12d93f4f3f99cf3c84bc32e78100c086122cc44R151.

pkliczewski (approved):

internal/workload/manager.go
```diff
@@ -175,13 +191,13 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
...
-			if err != nil {
-				if !os.IsNotExist(err) {
-					errors = multierror.Append(errors, fmt.Errorf("cannot remove existing manifest workload: %s", err))
-				}
+			if err := deleteFile(manifestPath); err != nil {
+				errors = multierror.Append(errors, fmt.Errorf("cannot remove existing manifest workload: %s", err))
+			}
+			authPath := w.getAuthFilePath(name)
+			if err := deleteFile(authPath); err != nil {
+				errors = multierror.Append(errors, fmt.Errorf("cannot remove existing workload auth file: %s", err))
```
masayag, line 199: what happens if the manifest gets deleted and the secret not ? we won't be able to track this file that needs to be deleted IIUC.

Would a folder structure per workload will be easier to manage in that case ?

internal/workload/manager.go
```diff
@@ -326,6 +360,12 @@ func (w *WorkloadManager) Deregister() error {
 		log.Errorf("failed to delete manifests directory: %v", err)
 	}
 
+	err = w.deleteAuthDir()
+	if err != nil {
+		errors = multierror.Append(errors, fmt.Errorf("failed to delete auth directory: %v", err))
+		log.Errorf("failed to delete ath directory: %v", err)
```
masayag, line 366: s/ath/auth

internal/workload/manager.go
```diff
@@ -450,10 +502,36 @@ func (w *WorkloadManager) toPod(workload *models.Workload) (*v1.Pod, error) {
...
+func (w *WorkloadManager) getAuthFilePathIfExists(workloadName string) string {
+	authFilePath := w.getAuthFilePath(workloadName)
+	if _, err := os.Stat(authFilePath); err != nil {
+		return ""
+	}
+	return authFilePath
+}
+
+func (w *WorkloadManager) podAuthModified(authPath string, auth string) bool {
+	if _, err := os.Stat(authPath); err != nil {
```
masayag, line 526: Is there a need to check the error to read the file only for 'file not found' like ?

masayag (changes requested):

internal/workload/manager.go
```diff
@@ -138,24 +143,35 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
...
+		err = w.storeFile(manifestPath, podYaml)
 		if err != nil {
 			errors = multierror.Append(errors, fmt.Errorf(
 				"cannot store manifest for workload '%s': %s", workload.Name, err))
 			continue
 		}
 
+		authFilePath, err = w.manageAuthFile(authFilePath, authFile)
+		if err != nil {
+			errors = multierror.Append(errors, err)
```
eloycoto, line 164: I'll append some context message to this error.

internal/workload/manager.go
```diff
@@ -195,6 +211,19 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
 	return errors
 }
 
+func (w *WorkloadManager) manageAuthFile(authFilePath, authFile string) (string, error) {
```
eloycoto, line 214: I'll add a comment to this function, mainly why we need to delete the authFilePath

eloycoto (approved):

internal/workload/manager.go
jakub-dzon: I didn't think so - if there is some disk/fs problem, it will be handled here: https://github.com/jakub-dzon/k4e-device-worker/pull/53/files#diff-8be2736004b01f09b6ffef40a12d93f4f3f99cf3c84bc32e78100c086122cc44R162

internal/workload/manager.go
jakub-dzon: Folder structure is a great idea for that.
I will make sure that the file is in expected state in the `ensureWorkloadsFromManifestsAreRunning` function and we can refactor the file structure in a separate PR - there would be many changes unrelated to this one.

PR #54, opened 2021-12-08
Add pod management via systemd

machacekondra:
This PR add pod/containers management via systemd instead of custom monitoring inside worker.

Currently we generate the systemd configuration for the pod&containers via `podman generate systemd` command. By default such configuration do restart of failed services only 5 times in 200ms and then exit the restart. So we need to provide the new config option to systemd which set the `RestartSec` option to service unit to appropriate value(default 15s).

We store the information about services persistently on filesystem in worker config directory in files `services.json`, so the info could survive restarts. The systemd units are generated in default location `/etc/systemd/system/`.

This feature will work only on RHEL8.5+, because of  [bug](https://bugzilla.redhat.com/show_bug.cgi?id=2030195).


internal/workload/wrapper.go
```diff
@@ -129,9 +140,65 @@ func (ww Workload) Run(workload *v1.Pod, manifestPath string, authFilePath strin
...
+	err = ww.serviceManager.Add(svc)
+	if err != nil {
+		return err
+	}
+
 	return ww.mappingRepository.Add(workload.Name, podIds[0])
 }
 
+func (ww Workload) removeService(workloadName string) error {
+	svc := ww.serviceManager.Get(workloadName + "_pod")
```
jakub-dzon, line 158: `workloadName + "_pod"` repeats in the code multiple times; WDYT about putting it in a function?

internal/workload/manager.go
```diff
@@ -266,86 +259,6 @@ func (w *WorkloadManager) toPodYaml(pod *v1.Pod) ([]byte, error) {
 	return podYaml, nil
 }
 
-func (w *WorkloadManager) ensureWorkloadsFromManifestsAreRunning() error {
```
jakub-dzon, line 269: With workload monitoring gone, please remove any notion of monitoring in tests; like here:https://github.com/jakub-dzon/k4e-device-worker/pull/54/files#diff-b99535cc9b7dcd200727082cda007cd8122ac6760506c4c0f825a9a1efc72b42R56

jakub-dzon (changes requested):
Please add tests for the new functionality.

internal/workload/service/systemd.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+func (s *Systemd) run(args []string) error {
+	args = append([]string{s.Path}, args...)
+	cmd := exec.Cmd{
+		Path: s.Path,
+		Args: args,
+	}
+
```
eloycoto, line 155: Should the log be added here? Like executing `%v`

eloycoto (approved):

internal/workload/wrapper.go
```diff
@@ -92,6 +99,10 @@ func (ww Workload) Remove(workloadName string) error {
 	if err := ww.workloads.Remove(id); err != nil {
 		return err
 	}
+	// Remove the service configuration from the system:
+	if err := ww.removeService(workloadName); err != nil {
```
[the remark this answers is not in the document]
machacekondra: That's a good question. You are right, that much safer would be to remove the service first, because it calls stop before and remove the system configuration later and so that correct way of doing things. Thanks for catching this!

internal/workload/service/systemd.go
```diff
@@ -0,0 +1,168 @@
...
+	mgr.lock.Lock()
+	defer mgr.lock.Unlock()
+
+	mgr.services[svc.Name] = svc
+
+	return mgr.write()
+}
+
+func (mgr *SystemdManager) Get(name string) *Systemd {
+	mgr.lock.Lock()
```
masayag, line 59: perhaps in this case `mgr.lock.RLock()` and `mgr.lock.RULock()` will be sufficient since we should allow multiple access for reading

internal/workload/service/systemd.go
```diff
@@ -0,0 +1,168 @@
...
+	for unit := range units {
+		unitNames = append(unitNames, unit)
+	}
+
+	return &Systemd{
+		Name:         name,
+		RestartSec:   DefaultRestartTimeout,
+		Path:         path,
+		Units:        unitNames,
+		UnitsContent: units,
```
masayag, line 102: Is there a need to define `Requires` section to list the dependencies? (is `podman.socket` a must or `nftables.service`) ?

masayag: Another question about dependencies, this is out of the scope for this PR since we don't yet support it, but for deployments-spec instead of pod-spec for the podman-kube-play, is there any relation between the generated service?

internal/workload/podman/podman.go
```diff
@@ -77,3 +86,13 @@ func (p *Podman) Start(workloadId string) error {
 	}
 	return nil
 }
+
+func (p *Podman) GenerateSystemdFiles(podName string, monitoringInterval int64) (map[string]string, error) {
+	useName := true
+	restartSec := (uint)(monitoringInterval)
```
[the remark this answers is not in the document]
machacekondra: Currently, there is no way to define unsinged int in swagger API: https://github.com/swagger-api/swagger-core/issues/3291

internal/workload/service/systemd.go
machacekondra: > Is there a need to define Requires section to list the dependencies? (is podman.socket a must or nftables.service) ?

`podman.socket` is a must for the `device-worker` itself, it won't start without it. Same for `nft`.
Podman pods, have it's own dependecies, which are generated from the podman code itself, like the containers, which pod consist of. But those are generated by the podman, no need to store it here. We just store all the units, which must be deleted, after the pod is removed.

> Another question about dependencies, this is out of the scope for this PR since we don't yet support it, but for deployments-spec instead of pod-spec for the podman-kube-play, is there any relation between the generated service?

We would just need to store, which services(units) must be removed in case deployment is removed. We have relation pod<->containers, so we would have relation deployment<->pods<->containers.

internal/workload/wrapper_test.go
```diff
@@ -0,0 +1,97 @@
...
+		wk = workload.NewWorkload(newPodman, netfilter, mappingRepository, serviceManager, 15)
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
+	})
+
+	Context("Systemd services", func() {
+
+		It("Create service if not exist", func() {
```
jakub-dzon, line 52: Please add 
```
// given
// when
// then
``` .
demarcation

pkliczewski (approved):

masayag (approved):

jakub-dzon (approved):

PR #55, opened 2021-12-08
support secrets

ydayagi:
receive collection of secrets required by the podSpecs
create/update the secrets in device

https://issues.redhat.com/browse/ECOPROJECT-221


go.mod
```diff
@@ -24,6 +24,7 @@ require (
 replace (
 	github.com/containers/buildah => github.com/containers/buildah v1.23.1
 	github.com/containers/common => github.com/containers/common v0.46.1-0.20211026130826-7abfd453c86f
+	github.com/jakub-dzon/k4e-operator => github.com/ydayagi/k4e-operator v0.0.0-20211208101641-8b430f6c0f5b
```
ydayagi, line 27: this will be removed after https://github.com/jakub-dzon/k4e-operator/pull/47 is merged

ydayagi: and then line 13 must be updated

pkliczewski (commented):
Should we cache secrets on the device with other manifests?

ydayagi:
> Should we cache secrets on the device with other manifests?

I don't see a reason to do that. We have a copy in memory. It comes from device configuration. As far as I understand the device configuration is saved in a file.

pkliczewski:
> 
> I don't see a reason to do that. We have a copy in memory. It comes from device configuration. As far as I understand the device configuration is saved in a file. 

I am thinking of use case when a device is disconnected from our control plane and it was rebooted. We need to make sure that device still can run the same workload.

ydayagi:
> > I don't see a reason to do that. We have a copy in memory. It comes from device configuration. As far as I understand the device configuration is saved in a file.
> 
> I am thinking of use case when a device is disconnected from our control plane and it was rebooted. We need to make sure that device still can run the same workload.

I will test it. But I think that podman secrets are persistent. For pods, I think Ondra handled it by converting pods to systemd units.

ydayagi:
> > > I don't see a reason to do that. We have a copy in memory. It comes from device configuration. As far as I understand the device configuration is saved in a file.
> > 
> > 
> > I am thinking of use case when a device is disconnected from our control plane and it was rebooted. We need to make sure that device still can run the same workload.
> 
> I will test it. But I think that podman secrets are persistent. For pods, I think Ondra handled it by converting pods to systemd units.

I checked. Secrets are still there after reboot.

internal/workload/manager.go
```diff
@@ -554,3 +557,29 @@ func (w *WorkloadManager) podAuthModified(authPath string, auth string) bool {
...
+
+func (w *WorkloadManager) updateSecrets(configuration models.DeviceConfigurationMessage) []error {
+	deviceSecrets, err := w.workloads.ListSecrets()
+	if err != nil {
+		return []error{err}
+	}
+	var errs []error
+	for _, configSecret := range configuration.Secrets {
+		if _, ok := deviceSecrets[configSecret.Name]; ok {
+			err = w.workloads.UpdateSecret(configSecret.Name, configSecret.Data)
```
machacekondra, line 569: If I am not mistaken, there will be big load to podman secret creation/deleteion, because you re-create the secret each time the `updateSecret` is called, which would be every heartbeat. Could we somehow check if there is change and re-create the secret only if there is change and don't re-create the secret always?

masayag (changes requested):

internal/configuration/configuration_test.go
```diff
@@ -312,6 +312,166 @@ var _ = Describe("Configuration", func() {
...
+			Expect(res).To(HaveLen(1))
+			Expect(res).To(Equal(cfg.Secrets))
+		})
+
+		It("When retreiving from config file", func() {
+			// given
+			cfg.Secrets = models.SecretList{
+				{
+					Name: "secret",
+					Data: "{}",
```
eloycoto, line 359: Shouldn't this be full of secrets?

internal/configuration/configuration_test.go
```diff
@@ -312,6 +312,166 @@ var _ = Describe("Configuration", func() {
...
+
+			// when
+			res := configManager.GetDeviceConfigurationMesage().Secrets
+
+			// then
+			Expect(res).To(HaveLen(1))
+			Expect(res).To(Equal(cfg.Secrets))
+		})
+
+		It("secrets equal", func() {
```
eloycoto, line 379: I'm lost on this Test, you defined two times the secret, but you not validated the output that it's the same, no? 
What is the result of the test here?

internal/configuration/configuration_test.go
```diff
@@ -312,6 +312,166 @@ var _ = Describe("Configuration", func() {
...
+			}
+
+			// when
+			err = configManager.Update(cfg)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("secrets not equal", func() {
```
eloycoto, line 427: The same as secrets are equal, you only test that are not equal, but you're not making any assertion on that one, no?

internal/workload/manager.go
```diff
@@ -118,6 +118,11 @@ func (w *WorkloadManager) Update(configuration models.DeviceConfigurationMessage
 		return errors
 	}
 
+	errs := w.updateSecrets(configuration)
+	if len(errs) != 0 {
+		errors = multierror.Append(errors, errs...)
```
eloycoto, line 123: Shouldn't return here? I mean, if secrets cannot be updated, the workloads should not do anything no? @jakub-dzon your input will be great here.

internal/workload/manager.go
eloycoto: +1

internal/workload/manager.go
```diff
@@ -539,3 +544,29 @@ func (w *WorkloadManager) podAuthModified(authPath string, auth string) bool {
 	}
 	return !bytes.Equal(file, []byte(auth))
 }
+
+func (w *WorkloadManager) updateSecrets(configuration models.DeviceConfigurationMessage) []error {
```
eloycoto, line 548: I need to say that here instead of models.DeviceConfigurationMessage, I would only send the secrets object instead of all config.

internal/workload/manager.go
eloycoto: You marked as resolved, but no comment or fixes

jakub-dzon: IMO "best effort" approach is good here; there might be workloads that don't need secrets and can be updated/started/removed.

eloycoto: make sense :+1:

internal/configuration/configuration.go
```diff
@@ -154,3 +159,21 @@ func (m *Manager) Deregister() error {
 	}
 	return nil
 }
+
+func isEqualUnorderedSecretLists(X models.SecretList, Y models.SecretList) bool {
```
jakub-dzon, line 163: Please use lower case for argument names

internal/configuration/configuration_test.go
jakub-dzon: A Saucerful of Secrets?;)

internal/workload/podman/podman.go
```diff
@@ -77,3 +79,32 @@ func (p *Podman) Start(workloadId string) error {
 	}
 	return nil
 }
+
+func (p *Podman) ListSecrets() (map[string]interface{}, error) {
```
jakub-dzon, line 83: Why is the value `interface{}` here?

jakub-dzon (changes requested):

internal/workload/podman/podman.go
jakub-dzon: OK. `map[string]struct{}` would be more memory-optimal and not so confusing - currently the values in the ma are always `nil` (https://github.com/jakub-dzon/k4e-device-worker/pull/55/files#diff-a98fd4c429032f68a1a6597b03c3ccba5f014918e82f41108aa962180f36e13eR90).

ydayagi: changed

jakub-dzon (approved):

masayag (approved):

PR #58, opened 2021-12-10
Add context to each log message

gciavarrini:
Signed-off-by: Gloria Ciavarrini gciavarrini@redhat.com

### DIFF
--- a/cmd/device-worker/main.go
+++ b/cmd/device-worker/main.go
@@ -43,12 +43,12 @@ func main() {
 	// Get initialization values from the environment.
 	yggdDispatchSocketAddr, ok = os.LookupEnv("YGG_SOCKET_ADDR")
 	if !ok {
-		log.Fatal("Missing YGG_SOCKET_ADDR environment variable")
+		log.Fatal("missing YGG_SOCKET_ADDR environment variable")
 	}
 
 	baseDataDir, ok := os.LookupEnv("BASE_DATA_DIR")
 	if !ok {
-		log.Warnf("Missing BASE_DATA_DIR environment variable. Using default: %s", defaultDataDir)
+		log.Warnf("missing BASE_DATA_DIR environment variable. Using default: %s", defaultDataDir)
 		baseDataDir = defaultDataDir
 	}
 
@@ -76,7 +76,7 @@ func main() {
 	// Listen on the provided socket address.
 	l, err := net.Listen("unix", r.GetAddress())
 	if err != nil {
-		log.Fatalf("Cannot start listening on %s err: %v", r.GetAddress(), err)
+		log.Fatalf("cannot start listening on %s err: %v", r.GetAddress(), err)
 	}
 
 	// Register as a Worker service with gRPC and start accepting connections.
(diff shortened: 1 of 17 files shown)


cmd/device-worker/main.go
```diff
@@ -13,14 +15,15 @@ import (
...
+	"log"
 	"net"
 	"os"
 	"path"
 	"time"
 
-	"git.sr.ht/~spc/go-log"
-
 	pb "github.com/redhatinsights/yggdrasil/protocol"
+	"go.uber.org/zap"
```
pkliczewski, line 25: What is the advantage to use this logger?

gciavarrini: I thought about using zap to allow a homogeneous log management for `k4e-operator` and `k4e-device-worker`.
Using different logs implies a different interface for the user in terms of verbosity levels.
Also, if I'm not wrong, go-log doesn't implement the logr.Logger interface, so we don't have the `WithValues` method to easily add context information to each log message in each involved project.
What do you think about that?

masayag: I'd be interesting to learn how does it affect the output log file.
The output goes to the file which is created by yggd iiuc, and it should contain the output for yggdrasil and its workers.
Will the file have a mixed format and verbosity ?

gciavarrini: Without my PR:
if `yggd` is started with `--log-level error` and `k4e-worker` with default log level (`info`) then the output will be something like
```
[yggd] 2021/12/13 18:07:48 [/usr/local/libexec/yggdrasil/device-worker] time="2021-12-13T18:07:48+01:00" level=info msg=......
[yggd] 2021/12/13 18:07:48 [/usr/local/libexec/yggdrasil/device-worker] time="2021-12-13T18:07:48+01:00" level=warning msg= .......
```
so yggd prints his own error logs + k4e-worker all logs (info, warn and error).
In this case there is a mix in term of verbosity.

with my PR there is also a mix in term of format

```[yggd] 2021/12/13 17:49:12 listening on socket: @yggd-dispatcher-nyfcMI
[yggd] 2021/12/13 17:49:12 [/usr/local/libexec/yggdrasil/device-worker] {"level":"info","ts":1639414152.9651952,"logger":"device-worker","caller":"device-worker/main.go:65","msg":"....}
[yggd] 2021/12/13 17:49:12 worker registered: {pid:146390 handler:device addr:@ygg-device-YkeGUn features:map[] detachedContent:false}
[yggd] 2021/12/13 17:49:12 [/usr/local/libexec/yggdrasil/device-worker] {"level":"info","ts":1639414152.966989,"logger":"device-worker","caller":"device-worker/main.go:102","msg":"......"}
```
Did I address properly your doubt?

So probably it's better to keep log management homogeneous between `yggd` and `k4e-device-worker`. Right?

masayag: yes, that was exactly my concern. since the entire output goes into the same log file, I'd prefer it to be the same format.

chenyosef:
General comments - can we try having some convention for our log messages? 
For example, I saw just in this PR messages start with "Cannot", "Can't", "cannot".
Some messages starts with capital letter, other don't

gciavarrini:
> General comment - can we try having some convention for our log messages? For example, I saw just in this PR messages start with "Cannot", "Can't", "cannot". Some messages starts with capital letter, other don't

Yes, I had noticed that too. I don't know what the convention is for this project but I actually think it's useful to standardize the log style.

pkliczewski:
Let's start with consistency so let's make them all alike.

masayag:
> 

Can you check if there is a convention across the yggdrasil project? after all we're all going to the same file

masayag (approved):

eloycoto (approved):

PR #60, opened 2021-12-22
Workload: Podman report containers IPS

eloycoto:
When creating a pod using kube play, the containers names, ids, or IPS
are not returned to the workload.

Sometimes the pod containers information is needed to be able to manage
anything extra, like metrics. So having a way to retrieve containers IPS
is helpful for other components.

This Change will call WorkloadStarted observer callback with the
following information:

```
[
  {
    "Id": "23c9d5820abbf9680e46c738b0753aec227269744bcbe79cbce98711a1222f4c",
    "Name": "",
    "Containers": [
      {
        "IPAddress": "10.88.0.41",
        "Id": "9455826dfaeacd69b61505d97a783c012190403e1558bb14d07a25fb76c31dce",
        "Name": "random-deployment-random-server"
      }
    ]
  }
]
```

Due to this is going to live in RAM for the whole process, callback will
only provide a minimal amount of data.


internal/workload/podman/podman.go
```diff
@@ -55,17 +74,46 @@ func (p *Podman) Remove(workloadId string) error {
...
 	}
-	var podIds []string
-	for _, pod := range report.Pods {
-		podIds = append(podIds, pod.ID)
+
+	var podIds = make([]*PodReport, len(report.Pods))
+	for i, pod := range report.Pods {
+		report := &PodReport{Id: pod.ID}
+		for _, container := range pod.Containers {
+			c, err := p.getContainerDetails(container)
```
eloycoto, line 109: Yes, there is no way to get the list of all containers with IPS in a single call :-(

internal/datatransfer/monitor.go
```diff
@@ -68,6 +69,10 @@ func (m *Monitor) GetLastSuccessfulSyncTime(workloadName string) *time.Time {
 	return nil
 }
 
+func (m *Monitor) WorkloadStarted(workloadName string, report []*podman.PodReport) {
```
gciavarrini, line 72: Can you explain why this no-op function is necessary? Why the parameters are needed?

internal/workload/podman/podman.go
```diff
@@ -69,17 +88,46 @@ func (p *podman) Remove(workloadId string) error {
...
+// getContainerDetails returns some basic information about a container
+func (p *podman) getContainerDetails(containerId string) (*ContainerReport, error) {
+	data, err := containers.Inspect(p.podmanConnection, containerId, nil)
+	if err != nil {
+		return nil, err
+	}
+	// this is the default one
+	network, ok := data.NetworkSettings.Networks["podman"]
+	if !ok {
+		return nil, fmt.Errorf("Cannot retrieve container '%s' network information", containerId)
```
gciavarrini, line 100: According to Staticcheck, error strings should not be capitalized

internal/workload/podman/podman.go
```diff
@@ -69,17 +88,46 @@ func (p *podman) Remove(workloadId string) error {
...
-	for _, pod := range report.Pods {
-		podIds = append(podIds, pod.ID)
+
+	var podIds = make([]*PodReport, len(report.Pods))
+	for i, pod := range report.Pods {
+		report := &PodReport{Id: pod.ID}
+		for _, container := range pod.Containers {
+			c, err := p.getContainerDetails(container)
+			if err != nil {
+				log.Errorf("Cannot get container information: %v", err)
```
gciavarrini, line 125: It is better to make the format of the log messages consistent in the project avoiding capitalize string.

jakub-dzon (approved):

gciavarrini (approved):

pkliczewski (approved):

internal/workload/wrapper.go
```diff
@@ -133,7 +134,12 @@ func (ww Workload) Run(workload *v1.Pod, manifestPath string, authFilePath strin
 	if err != nil {
 		return err
 	}
-	return ww.mappingRepository.Add(workload.Name, podIds[0])
+
+	for _, observer := range ww.observers {
+		observer.WorkloadStarted(workload.Name, podIds)
```
masayag, line 139: Will we be able to catch changes to the underlying containers if they were restarted by podman?

eloycoto: This is also raised by @jakub-dzon if the connection is not routable it'll fetch again the container IPs.

PR #61, opened 2021-12-23
Tooling: Running --race options.

eloycoto:
When running --race options some race-conditions can be found in the
system. This commits adds two things:

1) Tests with --race conditions and their fixes.
2) make build-debug and make install-debug to run device-worker with
   race conditions to be able to detect this cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 VERSION = 1.0
 RELEASE = 1
 DIST_DIR = $(shell pwd)/dist
-
+CGO_ENABLED = 0
 OS :=$(shell awk -F= '/^ID/{print $$2}' /etc/os-release)
 
 ifeq ($(OS),fedora)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -66,16 +66,25 @@ clean: ## Clean project
...
 
+build-debug: ## Build with race conditions and lock checker
+build-debug: BUILD_OPTIONS=--race
+build-debug: CGO_ENABLED=1
+build-debug: build
+
 build: ## Build device worker
 	mkdir -p ./bin
-	CGO_ENABLED=0 go build -o ./bin ./cmd/device-worker
+	CGO_ENABLED=$(CGO_ENABLED) go build $(BUILD_OPTIONS) -o ./bin ./cmd/device-worker
```
eloycoto, line 76: CGO_ENABLED is only for race, if not cannot be enabled.

gciavarrini (approved):
lgtm

jakub-dzon (approved):

PR #62, opened 2021-12-23
TSDB metrics storage wrapper

jakub-dzon:
This PR introduces wrapped Prometheus' TSDB as a solution for storing metrics on-device.

Metrics retention is controlled by https://github.com/jakub-dzon/k4e-operator/blob/d9c0ae0f8f3b4a756077a3384aba6f447a54abf1/api/v1alpha1/edgedevice_types.go#L43 .


internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+		appliedOptions: opts,
+	}, err
+}
+
+func (t *TSDB) Deregister() error {
+	metricsDir := t.db.Dir()
+	t.dbLock.Lock()
+	defer t.dbLock.Unlock()
+
+	err := t.db.Close()
```
ydayagi, line 72: shouldn't this be called every time worker stops/terminates/shutsdown?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	return toSeries(seriesSet), nil
+}
+
+func toSeries(seriesSet storage.SeriesSet) []Series {
+	var allSeries []Series
+	for {
+		if !seriesSet.Next() {
+			break
+		}
+		s := seriesSet.At()
```
ydayagi, line 118: the seriesSet is sorted. our map is not. that's ok?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	return appender.Commit()
+}
+
+func (t *TSDB) Update(config models.DeviceConfigurationMessage) error {
+	metricsConfiguration := config.Configuration.Metrics
+	maxBytes := DefaultMaxBytes
+	retentionDuration := DefaultRetentionDuration / time.Millisecond
+	if metricsConfiguration != nil && metricsConfiguration.Retention != nil {
+		maxBytes = int64(metricsConfiguration.Retention.MaxMib) * 1024 * 1024
+		retentionDuration = time.Duration(metricsConfiguration.Retention.MaxHours) * time.Hour / time.Millisecond
```
ydayagi, line 152: I don't think it is OK to set the duration directly with a number. Although Duration is an int64 it is supposed to be in nanoseconds and we set it to hours. We then use the fact that it is an int64 for converting to nanoseconds and then back to milliseconds.
I think we should not use the time package in this case.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil
+	}
+	log.Infof("Metrics retention changed. MaxBytes [%v -> %v] RetentionDuration: [%d -> %d]", t.appliedOptions.MaxBytes, maxBytes, t.appliedOptions.RetentionDuration, retentionDuration)
+
+	t.dbLock.Lock()
+	defer t.dbLock.Unlock()
+
+	dir := t.db.Dir()
+
+	err := t.db.Close()
```
ydayagi, line 164: we are closing the DB while someone else might be writing to it. For example, AddMetric

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	Time  int64
+	Value float64
+}
+type Series struct {
+	Labels     map[string]string
+	DataPoints []DataPoint
+}
+
+//go:generate mockgen -package=metrics -destination=storage_mock.go . Storage
+type Storage interface {
```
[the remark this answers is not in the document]
jakub-dzon: Because the package is called `metrics` it is not recommended to repeat that as part of type names inside the package.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, err
+	}
+	return &TSDB{
+		db:             db,
+		dbLock:         sync.RWMutex{},
+		appliedOptions: opts,
+	}, err
+}
+
+func (t *TSDB) Deregister() error {
```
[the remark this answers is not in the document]
jakub-dzon: The lock ensures that changes made by `Update` are visible to other go routines (which may be adding or reading metrics).

internal/metrics/metrics.go
jakub-dzon: You're right; the Read lock is held not long enough and probably should span throughout the whole `AddMetric` and `GetMetricsForTimeRange` methods.

internal/metrics/metrics.go
jakub-dzon: IMO for labels it's OK not to be sorted as order of labels doesn't have any significance.

cmd/device-worker/main.go
```diff
@@ -106,12 +107,19 @@ func main() {
 	wl.RegisterObserver(dataMonitor)
 	configManager.RegisterObserver(dataMonitor)
 	dataMonitor.Start()
+
+	metricsStore, err := metrics.NewTSDB(dataDir)
+	configManager.RegisterObserver(metricsStore)
+
+	if err != nil {
```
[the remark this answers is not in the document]
jakub-dzon: Kind of - if the device doesn't start before it's registered, it won't be seen as a CR, if it doesn't start after it has been registered and rebooted the last seen time in the device status wouldn't be updated.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	matchers := make([]*labels.Matcher, 0)
+	for _, lbl := range lbls {
+		matchers = append(matchers, labels.MustNewMatcher(labels.MatchRegexp, lbl, ".*"))
+	}
+	return matchers, nil
+}
+
+func toDbTime(t time.Time) int64 {
+	return t.UnixNano() / int64(time.Millisecond)
```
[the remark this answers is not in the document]
jakub-dzon: I could if we used go 1.17 or newe - it's not there in 1.16.

We can update go in a separate PR,

internal/metrics/metrics.go
jakub-dzon: You're right - I will change it.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func NewTSDB(dataDir string) (*TSDB, error) {
+	metricsDir := path.Join(dataDir, "metrics")
+	if err := os.MkdirAll(metricsDir, 0755); err != nil {
+		return nil, fmt.Errorf("cannot create directory: %w", err)
+	}
+	opts := tsdb.DefaultOptions()
+
+	opts.RetentionDuration = int64(DefaultRetentionDuration / time.Millisecond)
```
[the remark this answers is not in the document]
jakub-dzon: Done.

internal/metrics/metrics.go
jakub-dzon: Fixed - added signal handler in main.

internal/metrics/metrics.go
jakub-dzon: Done.

internal/metrics/metrics.go
jakub-dzon: Fixed.

cmd/device-worker/main.go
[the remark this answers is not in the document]
ydayagi: does the error here cause worker to fail? it seems we just log. I thought ondra added a way to report errors back to operator.

internal/metrics/metrics.go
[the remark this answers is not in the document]
ydayagi: The Deregister does not happen in Update, correct? I expect 'deregister' to wait for all routines that are using the metrics feature to stop (or at least stop using the feature). Otherwise, we close the DB while they are reading/writing.
In general, I would expect the deregister of the device to stop all routines and then do the termination/close actions.

ydayagi (approved):

internal/metrics/metrics.go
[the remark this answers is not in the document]
masayag: in that case, perhaps API will be more precise, since it will be referenced as `metrics.API`.

cmd/device-worker/main.go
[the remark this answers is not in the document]
jakub-dzon: Yes - the trick is that `log.Fatal` logs _and_ terminates the application.
The events are sent as part of the heartbeats and if the worker is not initialized, they wouldn't be sent (and event if, that would cause the last time seen be updated on the edge device CR - which is not true, because the worker is crashing).

internal/metrics/metrics.go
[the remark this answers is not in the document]
jakub-dzon: Yes, `Deregister` is not called in `Update`.
`Close` method, which is part of the `Deregister` operation waits for all the in-flight operations to end - it tries to take exclusive (write) lock which can only be taken if all read locks are released (taken by the other operations). 

As a side note - it actually doesn't matter if other goroutines complete their actions successfully - the metrics data is removed in the `Deregister` call.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,211 @@
...
+		allSeries = append(allSeries, Series{
+			Labels:     labelsMap,
+			DataPoints: dataPoints,
+		})
+	}
+	return allSeries
+}
+
+func (t *TSDB) AddMetric(value float64, labelsMap map[string]string) error {
+	t.dbLock.RLock()
```
masayag, line 152: only `RLock` is needed in this context? if we modify the state of the DB, the lock for writing should be more suitable.

masayag: Perhaps it will be better approach to use channels for submitting metrics instead of using lock to be non-blocking if that might have an impact on the workloads execution. In other words: Where can we find the caller for this method?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,211 @@
...
+	defer t.dbLock.RUnlock()
+
+	appender := t.db.Appender(context.Background())
+	_, err := appender.Append(0, labels.FromMap(labelsMap), toDbTime(time.Now()), value)
+	if err != nil {
+		return err
```
masayag, line 158: no need to handle a rollback in this case?

masayag (changes requested):

cmd/device-worker/main.go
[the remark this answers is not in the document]
ydayagi: wow. I didn't know that. thanks.

internal/metrics/metrics.go
jakub-dzon: There are no callers for that feature yet - we can re-think the approach, when the code that actually uses the DB is being written. Using channels looks to be a good idea.

Regarding the lock - it does not protect the content of the db, but the pointer to `TSDB` and it makes sure that the whatever it points to is visible and either completely usable or completely unusable.

internal/metrics/metrics.go
jakub-dzon: Good point! Probably would be safer to call it. Unfinished transaction might cause memory leaks...

cmd/device-worker/main.go
```diff
@@ -121,6 +131,20 @@ func main() {
...
+	signal.Notify(c, os.Interrupt, os.Kill, syscall.SIGTERM)
+	go func() {
+		select {
+		case sig := <-c:
+			log.Infof("Got %s signal. Aborting...\n", sig)
+			if err := metricsStore.Close(); err != nil {
+				log.Error(err)
+			}
+
+			os.Exit(0)
```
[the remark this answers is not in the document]
jakub-dzon: It's just an indication of a graceful shutdown.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,211 @@
...
+	"path"
+	"sync"
+	"time"
+)
+
+const (
+	millisInHour = 60 * 60 * 1000
+	// DefaultRetentionDuration specifies 7 days data retention
+	DefaultRetentionDuration = int64(7 * 24 * millisInHour)
+	// DefaultMaxBytes specifies no block data size limit
```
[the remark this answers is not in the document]
ydayagi: there's a story for monitoring device resources. https://issues.redhat.com/browse/ECOPROJECT-452

internal/metrics/metrics.go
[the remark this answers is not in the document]
ydayagi: The exclusive lock makes the routines wait. But it does not make them stop using the TSDB object. Immediately after Deregister releases the lock these routines will access TSDB object. Now 'TSDB.db' is not nil (perhaps close should set it to nil). But can it be used after closing? without panic? even if we can use it without panic I believe we will see an error in the log. It's not a big problem but it's noise in the log.

cmd/device-worker/main.go
```diff
@@ -121,6 +131,20 @@ func main() {
 		reg.RegisterDevice()
 	}
 
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt, os.Kill, syscall.SIGTERM)
```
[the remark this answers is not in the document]
jakub-dzon: Interesting question about `defer` - I am not sure if it would be always executed.

internal/metrics/metrics.go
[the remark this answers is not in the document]
jakub-dzon: Having db nil would require additional nil checks when we use it. When TSDB is closed and someone tries to commit, simply an error is returned - no panics.

masayag (approved):

internal/metrics/metrics.go
```diff
@@ -0,0 +1,214 @@
...
+
+	db, err := tsdb.Open(metricsDir, nil, nil, opts, nil)
+	if err != nil {
+		return nil, err
+	}
+	return &TSDB{
+		db:             db,
+		dbLock:         sync.RWMutex{},
+		appliedOptions: opts,
+	}, err
```
pkliczewski, line 67: At this point `err` will be nil. Why do we want to return it?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (t *TSDB) Deregister() error {
+	metricsDir := t.db.Dir()
+	err := t.Close()
+	if err != nil {
+		return err
```
pkliczewski, line 74: We log error in other place in this method. Why not to be consistent and log everywhere?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (t *TSDB) GetMetricsForTimeRange(tMin time.Time, tMax time.Time) ([]Series, error) {
+	log.Tracef("Getting metrics for %v - %v", tMin, tMax)
+
+	t.dbLock.RLock()
+	defer t.dbLock.RUnlock()
+
+	q, err := t.db.Querier(nil, toDbTime(tMin), toDbTime(tMax))
+	if err != nil {
+		return nil, err
```
pkliczewski, line 105: should we log here as well?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,214 @@
...
+	defer func() {
+		err = q.Close()
+		if err != nil {
+			log.Error(err)
+		}
+	}()
+
+	matchers, err := getAllLabelsMatchers(q)
+	if err != nil {
+		return nil, err
```
pkliczewski, line 116: or here?

pkliczewski (commented):
Is there any significant impact of this change on memory footprint of the worker?

internal/metrics/metrics.go
jakub-dzon: Good catch. That's a leftover from different implementation.

jakub-dzon:
@pkliczewski Base memory TSDB overhead is pretty low - 3MB; it will grow with the amount of data points collected.

pkliczewski (approved):

PR #64, opened 2022-01-06
Create a netfilter interface so we can mock it

machacekondra:
This PR add Netfilter interface, so we can mock it to properly test: https://github.com/jakub-dzon/k4e-device-worker/pull/54


machacekondra:
@jakub-dzon @masayag @eloycoto  PTAL

masayag (approved):

PR #65, opened 2022-01-06
Create a podman interface so we can mock it

machacekondra:
This PR add podman interface, so we can mock it to properly test: https://github.com/jakub-dzon/k4e-device-worker/pull/54


internal/workload/podman/podman.go
```diff
@@ -11,21 +11,32 @@ import (
 	api2 "github.com/jakub-dzon/k4e-device-worker/internal/workload/api"
 )
 
-type Podman struct {
+type Podman interface {
```
masayag, line 14: no need for `//go:generate mockgen...` ?

internal/workload/wrapper.go
```diff
@@ -40,8 +40,8 @@ type WorkloadWrapper interface {
 
 // Workload manages the workload and its configuration on the device
 type Workload struct {
-	workloads         *podman.Podman
-	netfilter         network.Netfilter
+	workloads         podman.Podman
+	netfilter         *network.Netfilter
```
masayag, line 44: this will be a pointer to an interface

machacekondra: Thanks! Wrong rebase.

internal/workload/podman/podman.go
machacekondra: Added, thanks.

masayag (approved):

PR #66, opened 2022-01-06
Create a mappingrepository interface so we can mock it

machacekondra:
This PR add mappingrepository interface, so we can mock it to properly test: #54


machacekondra:
@masayag PTAL

masayag (approved):

PR #67, opened 2022-01-10
Metrics scraper

jakub-dzon:
This PR introduces `metrics.HTTPScraper` type that is responsible for calling given HTTP endpoint and scraping (retrieving and parsing) metrics in openmetrics format exposed by it.

### DIFF
--- a/internal/metrics/scraper.go
+++ b/internal/metrics/scraper.go
@@ -1,0 +1,98 @@
+package metrics
+
+import (
+	"bufio"
+	"compress/gzip"
+	"context"
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
+	"io"
+	"net/http"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


eloycoto (approved):

machacekondra:
Prometheus SDK is not usable here?

pkliczewski (approved):

internal/metrics/scraper.go
```diff
@@ -0,0 +1,98 @@
...
+	"io"
+	"net/http"
+)
+
+const (
+	acceptHeader = `application/openmetrics-text; version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`
+)
+
+var (
+	UserAgent = fmt.Sprintf("k4e-device-worker")
```
masayag, line 20: why not a simple string assignment?

jakub-dzon: Good one! Blind copy&pase&edit...

jakub-dzon:
@machacekondra I've only found instrumentation SDK and the closest to what we need was prometheus-commons, which provides openmetrics parser.

masayag (approved):

PR #68, opened 2022-01-14
Metrics: Collect metrics from worker

eloycoto:
Hi, 

Three commits:

- Operator dependency version bump. 
- Podman Events commit to retrieve some config. 
- Workload metrics collect and added to the TSDB. 

Each commit describes what it's achieved.

Best regards


internal/metrics/daemon.go
```diff
@@ -0,0 +1,201 @@
...
+func (tg *TargetMetric) Stop() {
+	tg.lock.Lock()
+	defer tg.lock.Unlock()
+
+	tg.cancel()
+	tg.cancel = nil
+	return
+}
+
+func (tg *TargetMetric) createScrapper() {
```
jakub-dzon, line 90: `Scrapper` -> `Scraper`

internal/metrics/daemon.go
```diff
@@ -0,0 +1,201 @@
...
+			if err != nil {
+				log.Errorf("cannot store vector: %v", err)
+			}
+		}
+	}
+	go target.Start(cb)
+}
+
+// AddTarget adds a target to  listen for the  given workloadName.
+func (md *metricsDaemon) AddTarget(workloadName string, urls []string, interval time.Duration, labelsMap map[string]string) {
```
jakub-dzon, line 170: WDYT about changing `workloadName` to something more generic, like `targetName`? I would like to add system metrics target, which wouldn't have a "workload name".

internal/metrics/daemon.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			log.Errorf("cannot start scrapper for %v: %v", url, err)
+		}
+		scrappers = append(scrappers, scrappe)
+	}
+	tg.scrappers = scrappers
+}
+
+func (tg *TargetMetric) Run(ctx context.Context) []model.Vector {
+	res := []model.Vector{}
```
jakub-dzon, line 104: WDYT about just appending to a `model.Vector` instead of having slice of de-facto slices?

internal/metrics/daemon.go
```diff
@@ -0,0 +1,202 @@
...
+		interval:     interval,
+		urls:         urls,
+		forcetrigger: make(chan bool),
+	}
+}
+func (tg *TargetMetric) ForceEvent() {
+	tg.forcetrigger <- true
+}
+
+func (tg *TargetMetric) Start(cb ScrappeCallback) {
```
jakub-dzon, line 42: Is the callback really needed? Can't the logic of scraping and adding to TSDB be a method on `TargetMetric`?

internal/metrics/daemon.go
```diff
@@ -0,0 +1,202 @@
...
+	tg.createScrapper()
+	tg.ctx, tg.cancel = context.WithCancel(context.Background())
+	ticker := time.NewTicker(tg.interval)
+	tg.lock.Unlock()
+
+	getdata := func() {
+		log.Debugf("ticker run for workload '%v'", tg.workloadName)
+		ctx, _ := context.WithTimeout(context.Background(), scrapeTimeout)
+		data := tg.Run(ctx)
+		cb(data, map[string]string{workloadMetric: tg.workloadName})
```
jakub-dzon, line 61: WDYT about changning it into something more generic, like `metricSource` that could be both workload name or "system"?
We will also need to add device id to each of the metrics, but that I think can be done when the metrics are sent to the control plane.

cmd/device-worker/main.go
```diff
@@ -97,11 +98,21 @@ func main() {
...
 	configManager := configuration2.NewConfigurationManager(dataDir)
 
+	// --- Client metrics configuration ---
+	metricsStore, err := metrics.NewTSDB(dataDir)
+	configManager.RegisterObserver(metricsStore)
+	// Metrics Daemon
+	metricsDaemon := metrics.NewMetricsDaemon(metricsStore)
+	workloadMetricWatcher := metrics.NewWorkloadMetrics(metricsDaemon)
+	configManager.RegisterObserver(workloadMetricWatcher)
+	go metricsDaemon.Start()
```
pkliczewski, line 108: Is there a need for clean stop? Should we handle termination signals?

internal/metrics/workload.go
```diff
@@ -0,0 +1,109 @@
...
+	return &WorkloadMetrics{daemon: daemon}
+}
+
+func (wrkM *WorkloadMetrics) getWorkload(workloadName string) *models.Workload {
+	wrkM.lock.Lock()
+	defer wrkM.lock.Unlock()
+	return wrkM.workloadConfig[workloadName]
+}
+
+func (wrkM *WorkloadMetrics) Update(config models.DeviceConfigurationMessage) error {
```
pkliczewski, line 35: Why do we need to return `error`? It seems we always return `nil`

internal/workload/podman/podman.go
```diff
@@ -107,6 +125,89 @@ func (p *podman) getContainerDetails(containerId string) (*ContainerReport, erro
...
+			if err != nil {
+				log.Errorf("Cannot get pod report for pod '%v': ", wrk.Name, err)
+				continue
+			}
+			event := &PodmanEvent{
+				WorkloadName: wrk.Name,
+				Event:        StartedContainer,
+				Report:       report,
+			}
+			go func() { events <- event }()
```
pkliczewski, line 150: Why do we need gorouting here?

cmd/device-worker/main.go
eloycoto: It's part of the Start function. If it's already started will  stop it!

internal/metrics/workload.go
eloycoto: It's part of the interface for `Observe()`

internal/metrics/daemon.go
eloycoto: yeah, it's reported to the daemon, so the callback or channel is needed.

jakub-dzon: Can't the whole logic that relies on this callback be in `TargetMetric`?

jakub-dzon (approved):
Nice! To me it is easier to understand the code now.

internal/metrics/daemon.go
eloycoto: done, fixed.

pkliczewski (approved):

gciavarrini (approved):
lgtm

PR #69, opened 2022-01-14
Execute Ansible playbook

gciavarrini:
This PR add the possibility to execute an Ansible playbook in the flotta-worker-device.

The goal is to reproduce the behavior of rhc-worker-playbook (https://github.com/RedHatInsights/rhc-worker-playbook).

This PR depends on  https://github.com/project-flotta/flotta-operator/pull/115


internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,155 @@
...
+type EventList struct {
+	Events []message.AnsibleRunnerJobEventYaml
+}
+
+func NewEventList() *EventList {
+	return &EventList{Events: []message.AnsibleRunnerJobEventYaml{}}
+}
+
+//	TODO: Implement this function
+func (el EventList) AddEvent(event message.AnsibleRunnerJobEventYaml) {
```
gciavarrini, line 28: This function should be equivalent to the function in [rhc-worker-playbook](github.com/RedHatInsights/rhc-worker-playbook/blob/acd51a1cab05322e38d9ce68113af61ddcc7de54/rhc_worker_playbook/server.py#L109).

I can't understand where the "ansible_facts" dictionary is defined. 
The messages schema are described in the playbook-dispatcher [schema folder](https://github.com/RedHatInsights/playbook-dispatcher/tree/master/schema) section, if I'm not mistaken.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,222 @@
...
+	eventList := dispatcher.NewEventList()
+	// required event for cloud connector
+	onStart := dispatcher.ExecutorOnStart(reqFields.crcDispatcherCorrelationId, "")
+	eventList.AddEvent(onStart)
+
+	playbook := &playbook.AnsiblePlaybookCmd{
+		Playbooks:         []string{playbookYamlFile},
+		ConnectionOptions: ansiblePlaybookConnectionOptions,
+		Options:           ansiblePlaybookOptions,
+		// Exec:              events.AddEvent, ??
```
gciavarrini, line 135: Is there a way to invoke `AddEvent` function every time an Ansible event is received? (Similary to [rhc-worker-playbook](https://github.com/RedHatInsights/rhc-worker-playbook/blob/acd51a1cab05322e38d9ce68113af61ddcc7de54/rhc_worker_playbook/server.py#L212) )

Makefile
```diff
@@ -53,6 +53,15 @@ test-coverage: test
...
 	git ls-files --others --ignored --exclude-standard | grep "coverprofile$$" | xargs rm
 
+generate-tools:
+ifeq (, $(shell which gojsonschema))
+	go get github.com/atombender/go-jsonschema/cmd/gojsonschema
+endif
+
+generate-messages: generate-tools 
+	mkdir -p internal/ansible/model/message/
+	gojsonschema --yaml-extension yaml -p message internal/ansible/schema/ansibleRunnerJobEvent.yaml > internal/ansible/model/message/runner-job-event-gen.go
```
machacekondra, line 63: Why not use `-o` instead of stoud out redirect `>`?

gciavarrini: You're right! Thanks!

internal/ansible/ansible_manager.go
pkliczewski: We could use an interface similar to this [class](https://github.com/RedHatInsights/rhc-worker-playbook/blob/acd51a1cab05322e38d9ce68113af61ddcc7de54/rhc_worker_playbook/server.py#L102) and later one send it [back](https://github.com/RedHatInsights/rhc-worker-playbook/blob/acd51a1cab05322e38d9ce68113af61ddcc7de54/rhc_worker_playbook/server.py#L66) to the operator. The question is how to expose it to the users.

internal/ansible/dispatcher/dispatcher_events.go
pkliczewski: @machacekondra can you help?

pkliczewski (commented):
It would be great if you could think about end to end tests we can implement once CI is done.

internal/ansible/ansible_manager.go
machacekondra: Well there is no way using the ansible-playbook command AFAIK, only using ansible-runner. The only way is to write custom [callback plugin](https://docs.ansible.com/ansible/latest/plugins/callback.html). Or integrate ansible-runner command inside the go-ansible.

gciavarrini: Thanks @machacekondra for your suggestion! 
I created a custom executor to handle the runner results. Next step is to customize di StdOutCallback. Let's see if I'll manage to make it work.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,222 @@
...
+		if errorCode == NotInstalled {
+			log.Warn("WARNING: The k4e-worker-device package requires the ansible package to be installed.")
+		}
+		// required event for cloud connector
+		onFailed := dispatcher.ExecutorOnFailed(reqFields.crcDispatcherCorrelationId, "", errorCode, fmt.Sprintf("%v", errorDetails))
+		eventList.AddEvent(onFailed)
+	}
+
+	// send the final message after playbook completed
+	log.Info("Playbook run complete.")
```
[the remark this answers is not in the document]
gciavarrini: Added a debug log message with the dump of the playbook.

internal/ansible/dispatcher/dispatcher_events.go
gciavarrini: For the moment I set `gather_facts` to true so i can try to mimic the behavior. 
Probably, there is something in the `insights-client` validation, but I still need to investigate it.

Now I know how to get the data structure which contains the `ansible_facts` (for reference `AnsiblePlaybookJSONResults` of the go-ansible library)

Do you agree with this approach?

internal/server/deviceserver.go
```diff
@@ -5,36 +5,47 @@ import (
...
 	"git.sr.ht/~spc/go-log"
+	"github.com/jakub-dzon/k4e-device-worker/internal/ansible"
 	configuration2 "github.com/jakub-dzon/k4e-device-worker/internal/configuration"
 	registration2 "github.com/jakub-dzon/k4e-device-worker/internal/registration"
 	"github.com/jakub-dzon/k4e-operator/models"
 	pb "github.com/redhatinsights/yggdrasil/protocol"
 )
 
-type deviceServer struct {
+type DeviceServer struct {
```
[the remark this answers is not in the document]
gciavarrini: No reason :) I'll fix it.

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -17,15 +16,36 @@ import (
...
 type EventList struct {
-	Events []message.AnsibleRunnerJobEventYaml
+	RunnerJobEvents []message.AnsibleRunnerJobEventYaml
+	Events          []results.AnsiblePlaybookJSONResultsPlay
 }
 
 func NewEventList() *EventList {
-	return &EventList{Events: []message.AnsibleRunnerJobEventYaml{}}
+	return &EventList{
+		RunnerJobEvents: []message.AnsibleRunnerJobEventYaml{},
```
gciavarrini, line 25: @machacekondra 
I'm not sure what to do here.
The `executor_on_start` and the `executor_on_failed` events are  `message.AnsibleRunnerJobEventYaml` type.
The ansible result events are `results.AnsiblePlaybookJSONResultsPlay`.

Am I correct? Is this a good way to replicate the rhc-worker behavior?

machacekondra: I think you would need to write custom callback like this [one](https://github.com/alibaba/ansible-provider-docs/blob/master/lib/ansible/plugins/callback/json.py). And then parse the output as JSON.

internal/ansible/ansible_executor.go
```diff
@@ -0,0 +1,272 @@
...
+
+			printChan <- line
+		}
+		done <- struct{}{}
+	}()
+
+	for {
+		select {
+		case line := <-printChan:
+			fmt.Fprintf(w, "%s\n", line)
```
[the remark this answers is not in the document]
machacekondra: I think that won't be possible to do it add hoc, for ansible-playbook execution. Ad hoc access you have in the callback plugin [itself](https://github.com/alibaba/ansible-provider-docs/blob/master/lib/ansible/plugins/callback/json.py). So the only way is to create some json file which will be parsed after execution. Or create a plugin which will emit the events to here and parse it in runtime.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,222 @@
...
+				startTime = time.Now()
+			}
+
+		}
+	}
+
+	if err != nil {
+		// last event sould be the failure, find the reason
+		errorCode, errorDetails := parseFailure(eventList.Events[len(eventList.Events)-1])
+		if errorCode == NotInstalled {
```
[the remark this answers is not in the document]
machacekondra: It's fine to have it. I was just wondering about reasoning behind it.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,249 @@
...
+const (
+
+	//Required field names
+	crcDispatcherAttribute    = "crc_dispatcher_correlation_id"
+	returnURLAttribute        = "return_url"
+	responseIntervalAttribute = "response_interval"
+	defaultResponseInterval   = 300
+
+	// Failure types
+	NotInstalled = "ANSIBLE_PLAYBOOK_NOT_INSTALLED"
```
pkliczewski, line 50: What do you think about changing this to `ANSIBLE_NOT_INSTALLED`

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,249 @@
...
+		return fmt.Errorf("cannot create ansible playbook yaml file %s. Error: %v", playbookYamlFile, err)
+	}
+
+	/*
+	*******************************************
+	*  TODO : Verify playbook
+	*******************************************
+	 */
+
+	// eventList := dis
```
pkliczewski, line 143: please remove or uncomment

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,249 @@
...
+			log.Info("posting events...")
+			err := a.sendEvents(results, reqFields.returnURL, responseTo)
+			if err != nil {
+				log.Errorf("cannot post ansible playbook results: %v", err)
+			}
+		}
+	}
+
+	if errRunPlaybook != nil {
+		// last event sould be the failure, find the reason
```
pkliczewski, line 174: s/sould/should

pkliczewski (commented):
Please add end to end test in separate PR. @machacekondra can help with it

internal/ansible/ansible_manager.go
gciavarrini: It derives from [rhc-worker](https://github.com/RedHatInsights/rhc-worker-playbook/blob/8ddc5ccfc97290a021b4c4de673b92fedc38cbfb/rhc_worker_playbook/server.py#L86). 
But I agree, your suggestion makes the code clearer: I'll changed it.

pkliczewski:
We need to make sure ansible is part of our blueprint.

gciavarrini:
>

gciavarrini:
@pkliczewski 

> It would be great if you could think about end to end tests we can implement once CI is done.

> Please add end to end test in separate PR. @machacekondra can help with it

I created a new jira story to handle that. Is it ok?

gciavarrini:
> We need to make sure ansible is part of our blueprint.

Created a new Jira story for it.

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,479 @@
...
+			CrcDispatcherErrorDetails:  &errorDetails,
+		},
+	}
+}
+
+// Create the message with event data to send back to Dispatcher
+/*
+TODO: check if ComposeDispatcherMessage func is correct
+
+return Request('POST', return_url, files={
```
jakub-dzon, line 393: Please remove commented code.

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,479 @@
...
+
+// https://github.com/ansible/ansible-runner/blob/7e60c1f61ee6118f0b943e7897dc3a58ecc043bd/ansible_runner/display_callback/callback/awx_display.py#L477
+func (d *AnsibleDispatcher) PlaybookOnPlayStart(playbookFilename string, msgCounter int, play *results.AnsiblePlaybookJSONResultsPlay) *message.AnsibleRunnerJobEventYaml {
+	jobEvent := message.AnsibleRunnerJobEventYaml{
+		Event:       "playbook_on_play_start",
+		RunnerIdent: &d.deviceID,
+		// StartLine:   0, //Not available
+		// EndLine:     0, //Not available
+		Counter: msgCounter,
+		//Stdout:  "",                        //FIXME?
```
jakub-dzon, line 303: Please fix or remove.

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,479 @@
...
+	bodyContent, err := ioutil.ReadAll(httpReq.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.Data{
+		MessageId:  uuid.New().String(),
+		Directive:  returnURL,
+		ResponseTo: responseTo,
+		Metadata:   map[string]string{"Content-Type": "application/vnd.redhat.playbook.v1+jsonl"}, // TODO: is it correct? In python metadata=req.headers
```
jakub-dzon, line 420: Is it correct?

cmd/device-worker/main.go
```diff
@@ -145,8 +146,24 @@ func main() {
 
 	reg := registration2.NewRegistration(&hw, deviceOs, dispatcherClient, configManager, hbs, wl, dataMonitor, metricsStore, systemMetricsWatcher)
 
+	// defined how to connect to hosts
+	ansiblePlaybookConnectionOptions := &options.AnsibleConnectionOptions{ //FIXME
```
eloycoto, line 150: Is using ssh-keys to connect no? Should we add the ssh-keys for this?  or something?

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// required event for cloud connector
+	onStart := a.ansibleDispatcher.ExecutorOnStart(reqFields.crcDispatcherCorrelationID, "")
+	a.ansibleDispatcher.AddRunnerJobEvent(onStart)
+
+	executionCompleted := make(chan error)
+	playbookResults := make(chan *ansibleResults.AnsiblePlaybookJSONResults)
+
+	// execute
+	go execPlaybook(executionCompleted, playbookResults, playbookCmd, time.Duration(reqFields.responseInterval)*time.Second)
```
eloycoto, line 151: Should we have a context to stop this in some way? Worried that get stuck the playbook and there is no way to cancel, etc..

eloycoto: I mean, I'm worried that something waits forever on ansible, and playbook got stuck on device

internal/ansible/dispatcher/dispatcher_events_test.go
```diff
@@ -0,0 +1,352 @@
...
+						},
+					},
+				},
+			},
+		}
+	)
+	BeforeEach(func() {
+		ansibleDispatcher = dispatcher.NewAnsibleDispatcher("test-device-id-123457")
+	})
+	AfterEach(func() {})
```
eloycoto, line 59: I would delete this one.

eloycoto (commented):
Overall LGTM, I have more questions than issues :-)

cmd/device-worker/main.go
```diff
@@ -145,8 +146,24 @@ func main() {
...
+		Connection: "local",
+		User:       "test_user",
```
masayag, line 152: should we add this use to the blueprint so each edge device created will have a user name `test_user` ?

masayag:
pls update PR message

> This PR add the possibility to execute an Ansible playbook in the k4e-worker-device.

to

> This PR add the possibility to execute an Ansible playbook in the flotta-worker-device.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+	NotInstalled = "ANSIBLE_NOT_INSTALLED"
+	UndefError   = "UNDEFINED_ERROR"
+)
+
+var (
+	buff, buffErr *bytes.Buffer
+
+	playbookExecutor *execute.DefaultExecute
+	playbookCmd      *playbook.AnsiblePlaybookCmd
+	playbookYamlFile = ""
```
masayag, line 59: iiuc, we should be able to run several playbook in parallel.
meaning `HandlePlaybook` can be run concurrently.
If this is the case, we can't rely on global variables.

masayag (commented):
Is there an operator side for this to view the format of the request or how the swagger.yaml on operator side needs to be extended?

cmd/device-worker/main.go
machacekondra: I think that user is not needed for connection=local, is it?

cmd/device-worker/main.go
machacekondra: Connection local is not using ssh-keys.

cmd/device-worker/main.go
```diff
@@ -145,8 +146,24 @@ func main() {
...
+	}
+	// defined which should be the ansible-playbook execution behavior and where to find execution configuration.
+	ansiblePlaybookOptions := &playbook.AnsiblePlaybookOptions{ //FIXME
```
machacekondra, line 155: What exactly need to be fixed her?

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+
+	found := false
+	dataDir := "/tmp"
+
+	var err error
+	responseTo := d.MessageId
+	if payloadStr == "" {
+		return fmt.Errorf("missing playbook string in message")
+	}
+	if reqFields.crcDispatcherCorrelationID, found = d.GetMetadata()[crcDispatcherAttribute]; !found {
```
machacekondra, line 112: Would extracting the `d.GetMetadata()` to some variable make sense here?

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+		return MissingAttributeError(crcDispatcherAttribute, d.GetMetadata())
+	}
+	if responseInterval, found := d.GetMetadata()[responseIntervalAttribute]; !found {
+		log.Warnf(missingAttributeMsg(responseIntervalAttribute, d.GetMetadata())+" Defaulting to "+strconv.Itoa(defaultResponseInterval), d.GetMetadata())
```
machacekondra, line 116: isnt the second d.GetMetadata attribute not needed?

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+	}
+
+	if errRunPlaybook != nil {
+		// last event should be the failure, find the reason
+		msgList := a.ansibleDispatcher.GetMsgList()
+		errorCode, errorDetails := parseFailure(msgList[len(msgList)-1])
+		if errorCode == NotInstalled {
+			log.Warn("WARNING: The flotta-worker-device package requires the ansible package to be installed.")
```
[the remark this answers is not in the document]
machacekondra: s/flotta-worker-device/flotta-agent

cmd/device-worker/main.go
gciavarrini: I'm unsure about what it is the right configuration.
@machacekondra you're right, connection local doesn't require either ssh or username.

my doubts here are: can a playbook connect to remote host? Should the connection type be only local? Otherwise how to configure these parameters?

internal/ansible/ansible_manager.go
gciavarrini: Yes, you're right. I'll fix this.

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
gciavarrini: Do you think that adding ansible to out blueprint as [suggested](https://github.com/project-flotta/flotta-device-worker/pull/69#issuecomment-1046749656) by @pkliczewski isn't sufficient? 
In addition, @jakub-dzon has already suggested to split flotta agent and "playbook agent", when the first implementation will be completed. In that case, are the check on ansible support and prevent worker start still necessary?

cmd/device-worker/main.go
gciavarrini: I'm not sure which Fields should be set and what are the correct values.

cmd/device-worker/main.go
eloycoto: and, if the connection is on test_user, how can install a new package for example? Should we add that to sudo, no?

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
jakub-dzon: Maybe that's an occasion to implement some kind of "capabilities" mechanism, where our agent reports to the control plane what it supports? In that case the operator would inform the user that certain devices can't be targets of Ansible Playbooks.

Informing user about lack of ansible with heartbeat event seems like a good idea as long as we don't spam the control plane with same event.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,248 @@
...
+			log.Warnf(invalidAttributeMsg(responseIntervalAttribute, d.GetMetadata()[responseIntervalAttribute]+" Defaulting to "+strconv.Itoa(defaultResponseInterval)))
+			reqFields.responseInterval = defaultResponseInterval
+		}
+	}
+	if reqFields.returnURL, found = d.GetMetadata()[returnURLAttribute]; !found {
+		return fmt.Errorf(missingAttributeMsg(returnURLAttribute, d.GetMetadata()))
+	}
+
+	playbookYamlFile = path.Join(dataDir, "ansible_playbook_"+d.MessageId+".yml")
+	defer os.Remove(playbookYamlFile)
```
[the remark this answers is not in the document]
masayag: @gciavarrini pls change the order. the `defer` should be moved after the `if err != nil`

internal/server/deviceserver.go
```diff
@@ -15,18 +20,47 @@ type deviceServer struct {
...
+		ansibleManager:      ansibleManager,
 	}
 }
 
 // Send implements the "Send" method of the Worker gRPC service.
 func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error) {
 	go func() {
+		//check if it is an ansible playbook message
+
+		if x, found := d.GetMetadata()["ansible-playbook"]; found && x == "true" {
```
masayag, line 39: How do we make sure not to execute the same playbook over and over again after each call to `GetUpdates` is invoked? 
After all, if the playbook is part of device-configuration, we'll be sending that information few times per-minute which isn't desired to execute over and over (even if idempotent).
Also, which feedback will be sent back on successful or failed playbook execution?

masayag (changes requested):

internal/ansible/dispatcher/dispatcher_events.go
gciavarrini: No , it wasn't. I updated the code but I'm still investigating on how to retrieve the identity header field.

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
gciavarrini: argh! you're right!

internal/server/deviceserver.go
gciavarrini: > How do we make sure not to execute the same playbook over and over again 
The idea is that it's the operator's responsibility to know whether or not to send the playbook to the device. Makes sense for you?

>which feedback will be sent back on successful or failed playbook execution
The device will send the `ansible events` as results either to the operator or the redhat console ingress service.

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
pkliczewski: I think it is good idea but for now ansible playbook runner is part of our worker. Once we will decide to split it into 2 workers it makes sense to implement it. Yggdrasil seems like correct place to do it.

internal/server/deviceserver.go
masayag: I think we need to make sure to execute the same playbook twice in parallel.
This needs to be handled on the device side.

From declarative API perspective, the playbook will remain part of the CR.
However, the device-operator api may omit is as part of `get-updates` response if not needed.
And yet, there need to be a definition under which condition not to send a playbook that was already executed.
(reattempt on failure, not to resend a playbook that was completed successfully, assuming it wasn't changed).
This needs to be well defined and documented.

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
masayag: this wasn't addressed yet, therefore I'll keep this comment as "unresolve"

internal/server/deviceserver.go
```diff
@@ -55,3 +89,15 @@ func (s *deviceServer) NotifyEvent(ctx context.Context, in *pb.EventNotification
 	}
 	return &pb.EventReceipt{}, nil
 }
+
+func getTimeout(metadata map[string]string) time.Duration {
+	timeout := 300 * time.Second // Deafult timeout
+	if timeoutStr, found := metadata["ansible-playbook-timeout"]; found {
+		timeoutVal, err := strconv.Atoi(timeoutStr)
+		if err != nil {
```
masayag, line 97: If the timeout provided by the user is invalid, shouldn't it be logged?
On the other hand, we are expected to validate it as part of the API, therefore it makes sense we shouldn't fail here.
And yet, if the intention is to reuse this worker by other clients at some point and not only by the operator, a warning might be useful.

internal/ansible/ansible_manager.go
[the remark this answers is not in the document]
gciavarrini: sorry, something went wrong with the push. Now It should be addressed/

internal/server/deviceserver.go
gciavarrini: > I think we need to make sure to execute the same playbook twice in parallel.

@masayag Any suggestion on how to do this? An in memory map with playbook in execution is acceptable? Or there are other mechanisms for this in the device?

masayag: What will happen to the executed playbook if the device-worker process is died and restarted?
If it terminates the playbook, we might not have information about its completion.
Are there any on-disk resources created in this flow? Or any way to obtain playbook status outside of the device-worker?
The recommendation will be influenced by the answers to the above, since we'd might want to be able to report playbook execution status if it was cancelled prematurely or even to re-attempt running it.
It can be achieve by keeping track of it on disk, and reflect the status while not completed using in-memory map (since there isn't really need to read it from disk every time if nothing changed).

masayag: and to complete the other side, while we're running a playbook, we should dismiss the same playbook on the next 'get-updates' call.

gciavarrini: > Are there any on-disk resources created in this flow? 

No

> Or any way to obtain playbook status outside of the device-worker?

AFAIK, No. 

> and to complete the other side, while we're running a playbook, we should dismiss the same playbook on the next 'get-updates' call.

Is the `get-updates` call this https://github.com/project-flotta/flotta-operator/blob/389e8a025bb6525fb563f80d00c49ccb71f419b3/internal/yggdrasil/yggdrasil.go#L142 ?

masayag: yes, that's the one.

internal/ansible/ansible_manager.go
```diff
@@ -61,6 +62,11 @@ func NewAnsibleManager(
 		return nil, fmt.Errorf("ansible manager cannot initialize mapping repository: %w", err)
 	}
 
+	_, err = exec.LookPath("ansible")
```
eloycoto, line 65: maybe check two things here: 

- If Ansible version is supported. 
- If Ansible is executable for that user

pkliczewski: We can check ansible version but we are in control which version lands in the image.

gciavarrini: `ansible` should be installed via our blueprint. Since this is a "prototype", in my opinion, we can leave this check looser for the moment. WDYT?

pkliczewski:
@gciavarrini please rebase

internal/ansible/ansible_manager.go
masayag: +1. Since we own the blueprint (at least for now), we know what it must contain.

internal/server/deviceserver.go
gciavarrini: Added

Makefile
```diff
@@ -77,7 +86,7 @@ bump-operator: ## Bump flotta operator version dependency to the latest main
 
 clean: ## Clean project
 	go mod tidy
-	rm -rf bin
+	rm -rf bin internal/ansible/model
```
eloycoto, line 89: why the model should be deleted?

cmd/device-worker/main.go
```diff
@@ -162,17 +161,31 @@ func main() {
 		metricsStore,
 	)
 
+	dataDirPlaybook := path.Join(baseDataDir, "devicePlaybooks")
+	/* #nosec */
+	if err := os.MkdirAll(dataDirPlaybook, 0755); err != nil {
+		log.Fatal(fmt.Errorf("cannot create directory: %w", err))
```
eloycoto, line 167: `log.Fatalf`

cmd/device-worker/main.go
```diff
@@ -162,17 +161,31 @@ func main() {
...
+	}
+	ansibleManager, err := ansible.NewAnsibleManager(dispatcherClient, dataDirPlaybook)
+	if err != nil {
+		log.Fatalf("cannot start ansible manager, err: %v", err)
+	}
+	err = ansibleManager.ExecutePreviousPlaybooks()
+	if err != nil {
+		log.Fatalf("cannot run previous ansible playbooks, err: %v", err)
```
eloycoto, line 175: If playbook failed here, should be stop the device? I'm not sure about TBH

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+	"github.com/project-flotta/flotta-device-worker/internal/ansible/mapping"
+	"github.com/project-flotta/flotta-device-worker/internal/ansible/model/message"
+	"github.com/project-flotta/flotta-operator/models"
+
+	pb "github.com/redhatinsights/yggdrasil/protocol"
+)
+
+// AnsibleManager handle ansible playbook execution
+type AnsibleManager struct {
+	wg                *sync.WaitGroup
```
eloycoto, line 32: why a pointer here, maybe is not needed at all and we can avoid null exceptions, no?

internal/ansible/ansible_manager.go
eloycoto: Works for me :+1:

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+		MappingRepository: mappingRepository,
+	}, nil
+}
+
+func MissingAttributeError(attribute string, metadata map[string]string) error {
+	return fmt.Errorf(missingAttributeMsg(attribute, metadata))
+}
+
+func missingAttributeMsg(attribute string, metadata map[string]string) string {
+	return fmt.Sprintf("missing attribute %s in message metadata %v", attribute, metadata)
```
eloycoto, line 83: on writing maps I would use `%+v` directive.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+func (a *AnsibleManager) HandlePlaybook(playbookCmd *playbook.AnsiblePlaybookCmd, d *pb.Data, timeout time.Duration) error {
+	var err error
+	buffOut := new(bytes.Buffer)
+	buffErr := new(bytes.Buffer)
+
+	setupPlaybookCmd(playbookCmd, buffOut, buffErr)
+
+	deviceConfigurationMessage := models.DeviceConfigurationMessage{}
+
+	if string(d.Content) == "" {
```
eloycoto, line 106: this can be `len(d.cContent) == 0`, no? and no need to have the string.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+	playbookYamlFile := path.Join(dataDir, "ansible_playbook_"+d.MessageId+".yml")
+	err = os.WriteFile(playbookYamlFile, []byte(payloadStr), 0600)
+	if err != nil {
+		return fmt.Errorf("cannot create ansible playbook yaml file %s. Error: %v", playbookYamlFile, err)
+	}
+	defer os.Remove(playbookYamlFile)
+
+	/*
+	*******************************************
+	*  TODO : Verify playbook
```
eloycoto, line 141: Due to the changes, I would create a task and track here the ID.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+		case <-ctx.Done():
+			a.wg.Done()
+			return fmt.Errorf("execution timeout reached for playbook in messageID %s. Error: %v", d.MessageId, ctx.Err())
+		case errRunPlaybook = <-executionCompleted:
+			a.wg.Done()
+			if err != nil {
+				return fmt.Errorf("ansible playbook execution completed with error. [MessageID: %s, Error: %v]", d.MessageId, err)
+			}
+			log.Infof("ansible playbook execution completed of messageID %s", d.MessageId)
+			break loop
```
eloycoto, line 172: This loop definition is not needed at all, no?

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,351 @@
...
+		if res == nil && err != nil {
+			return err
+		}
+		buffOut.Reset()
+		buffErr.Reset()
+	}
+	return errors
+}
+
+func (a *AnsibleManager) StopPlaybooks() {
```
eloycoto, line 248: I'm not sure about this Stop function, should be waitPlaybooks finish or something like that, no?

internal/ansible/ansible_manager_test.go
```diff
@@ -0,0 +1,261 @@
...
+				Content:   msgContentByte,
+			}
+
+			//when
+			err = ansibleManager.HandlePlaybook(playbookCmd, &message, timeout)
+			Expect(err).To(HaveOccurred())
+			Expect(client.latestData).To(BeNil())
+		})
+	})
+	BeforeEach(func() {
```
eloycoto, line 191: I'm worried here, this BeforeEach/AfterEach are defined twice, maybe this should be inside of Context("on-restart"), no?

internal/ansible/dispatcher/dispatcher_events_test.go
```diff
@@ -0,0 +1,375 @@
...
+		//When
+		jsonBytes, err := dispatcher.CreateMessage(msgList)
+
+		//Expect
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(string(jsonBytes)).To(Equal(expectedJson))
+	})
+	It("Playbook on runner skipped", func() {
+		//Given
+		expectedJson :=
```
eloycoto, line 321: This json is very similar, and after that you do replace, a good thing is to use a go struct, have something like


`get_expected_json := func(UUID, counter) []bytes` and it's a bit clear, but keep like this, this PR is already big!

internal/ansible/mapping/mapping.go
```diff
@@ -0,0 +1,188 @@
...
+	}, nil
+}
+
+func (m *mappingRepository) GetAll() map[int]string {
+	all := make(map[int]string)
+	keys := make([]int64, 0, len(m.modTimeToPath))
+	for k := range m.modTimeToPath {
+		keys = append(keys, k)
+	}
+	// sort.Float64Slice(keys)
```
eloycoto, line 79: Just remove thease ones.

internal/server/deviceserver.go
```diff
@@ -39,6 +46,33 @@ func NewDeviceServer(configManager *configuration2.Manager, registrationManager
 // Send implements the "Send" method of the Worker gRPC service.
 func (s *deviceServer) Send(_ context.Context, d *pb.Data) (*pb.Response, error) {
 	go func() {
+		//check if it is an ansible playbook message
+
+		if x, found := d.GetMetadata()["ansible-playbook"]; found && x == "true" {
+			log.Debugf("Received message %s with 'ansible-playbook' metadata", d.MessageId)
+
+			// defined how to connect to hosts
+			ansiblePlaybookConnectionOptions := &options.AnsibleConnectionOptions{
```
eloycoto, line 55: TBH: I would move this to Ansible package and be like ansible.GetPlaybookCMD() that returns `playbook.AnsiblePlaybookCmd` it belongs more to the ansible package ;-)

eloycoto (commented):
Overall the PR is great, very consistent and almost all are nitpicks, great work Gloria :-)

cmd/device-worker/main.go
gciavarrini: My idea here is: if one of the playbooks fails then the device is in an inconsistent state. That's why in my opinion the device should stop. WDYT?

internal/ansible/ansible_manager.go
gciavarrini: I didn't add the Jira story ID because isn't accessible for non Red Hatters

internal/ansible/dispatcher/dispatcher_events_test.go
gciavarrini: I know it isn't ideal, but if even the differences among expectedJson strings are small... there are still there. 
I avoided factoring the code further hoping to make the test more talky. But maybe I didn't succeed!
Maybe we can improve this in another PR.

cmd/device-worker/main.go
eloycoto: @jakub-dzon ^^^

eloycoto:
LGTM, let's wait on @jakub-dzon  feedback on this one, but I'm happy to merge it as it is: 

https://github.com/project-flotta/flotta-device-worker/pull/69#discussion_r830130274

jakub-dzon:
:eyes:

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,369 @@
...
+		buffErr.Reset()
+	}
+	return errors
+}
+
+func (a *AnsibleManager) WaitPlaybookCompletion() {
+	a.wg.Wait()
+}
+
+func (a *AnsibleManager) GetPlakbookCommand() playbook.AnsiblePlaybookCmd {
```
jakub-dzon, line 253: `GetPlaybookCommand`

cmd/device-worker/main.go
gciavarrini: @masayag ATM is a single attempt. IMO, we can change this behavior once the playbook execution architecture will be defined, if necessary.

cmd/device-worker/main.go
```diff
@@ -162,17 +161,31 @@ func main() {
 		metricsStore,
 	)
 
+	dataDirPlaybook := path.Join(baseDataDir, "devicePlaybooks")
+	/* #nosec */
+	if err := os.MkdirAll(dataDirPlaybook, 0755); err != nil {
```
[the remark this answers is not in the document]
gciavarrini: I created the directory consistently with https://github.com/project-flotta/flotta-device-worker/blob/58fc496291e3b64bf8890f26b598c44e586ab9d3/cmd/device-worker/main.go#L93 

Do you think I should change both? If yes, should I remove `/* #nosec */` and use `0750` ? Or something else?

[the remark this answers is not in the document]
masayag: yes, `0750` should be sufficient for the device.

internal/ansible/mapping/mapping.go
```diff
@@ -0,0 +1,186 @@
...
+	h := sha256.New()
+	h.Write([]byte(fileContent))
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
+
+func (m *mappingRepository) Add(fileContent string, modTime time.Time) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	filePath := path.Join(m.configDir, m.GetSha256(fileContent))
+	err := os.WriteFile(filePath, []byte(fileContent), 0640) //#nosec
```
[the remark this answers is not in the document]
masayag: you can run `gosec ./...` and see if it complains or not.
I don't think we have anything specific to override the default values of files and directory permissions.
The defaults are 0755 and 0600
See https://github.com/securego/gosec/blob/f285d612b508abefc9c3dd85517aa11c41bb1c6f/rules/fileperms.go#L101
and https://github.com/securego/gosec/blob/f285d612b508abefc9c3dd85517aa11c41bb1c6f/rules/fileperms.go#L67
so in this case, if 0640 is needed, than `#nosec` should  remain.

cmd/device-worker/main.go
jakub-dzon: `log.Fatalf` should be changed to `log.Errorf`. Fatalf will call `Exit(0)` and I don't think it is desired.

gciavarrini: @jakub-dzon yes, I forgot to change it! Thanks!

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// execPlaybookSync executes the ansible playbook synchronously.
+// if error is not nil, then an error occured during the playbook execution (e.g. host unreachable), but this does't mean
+// that there are no results available. In other words, a successful call returns *ansibleResults.AnsiblePlaybookJSONResults not nil.
+func (a *Manager) execPlaybookSync(
+	playbook *playbook.AnsiblePlaybookCmd,
+	timeout time.Duration,
+	buffOut *bytes.Buffer,
+	mappingRepository mapping.MappingRepository) (*ansibleResults.AnsiblePlaybookJSONResults, error) {
```
jakub-dzon, line 353: `mappingRepository` is unused here, but I think the playbook should be removed from it. when the execution ends.

internal/ansible/ansible_manager.go
```diff
@@ -0,0 +1,390 @@
...
+	buffOut := new(bytes.Buffer)
+	buffErr := new(bytes.Buffer)
+
+	playbookCmd = setupPlaybookCmd(playbookCmd, buffOut, buffErr)
+	allPlaybooks := a.MappingRepository.GetAll()
+	var errors error
+	for _, v := range allPlaybooks {
+		playbookCmd.Playbooks = []string{v}
+		res, err := a.execPlaybookSync(&playbookCmd, timeout, buffOut, a.MappingRepository)
+		log.Error(err)
```
jakub-dzon, line 227: `if err != nil`

jakub-dzon (changes requested):

cmd/device-worker/main.go
```diff
@@ -162,17 +160,30 @@ func main() {
 		metricsStore,
 	)
 
+	dataDirPlaybook := path.Join(baseDataDir, "devicePlaybooks")
+	if err := os.MkdirAll(dataDirPlaybook, 0750); err != nil {
+		log.Fatalf("cannot create directory: %v", err)
+	}
+	ansibleManager, err := ansible.NewAnsibleManager(dispatcherClient, dataDirPlaybook)
+	if err != nil {
+		log.Fatalf("cannot start ansible manager, err: %v", err)
```
machacekondra, line 169: Do we consider Ansible as hard dependency, so without Ansible we can't use normal workloads? Or should it be soft, so if Ansible is not installed, we still can use device-worker, but without Ansible the Playbooks CRs will fail to execute?

gciavarrini: I added this because when there wasn't `ansible` in gitlab ci the test in my PR panicked (https://github.com/project-flotta/flotta-device-worker/runs/5557779128?check_suite_focus=true#step:5:18). 
I see your point. Let me fix it.

jakub-dzon (approved):

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,411 @@
...
+	msgCounter int,
+	host *string,
+	tasks *results.AnsiblePlaybookJSONResultsPlayTaskItem,
+	hosts *results.AnsiblePlaybookJSONResultsPlayTaskHostsItem,
+	taskUUIDs map[string]int,
+	duplicateTaskCount map[string]int) *message.AnsibleRunnerJobEventYaml {
+	taskUUID := tasks.Id
+	if occurence, found := taskUUIDs[taskUUID]; found {
+		if occurence > 1 {
+			// 	 When this task UUID repeats, it means the play is using the
```
machacekondra, line 255: Don't we always run on localhost so only with one host? It's fine to have this functionality implemented, but I think it won't ever be used in our case.

internal/ansible/dispatcher/dispatcher_events.go
```diff
@@ -0,0 +1,411 @@
...
+
+func (d *AnsibleDispatcher) RunnerOnSkipped(
+	runnerID string,
+	playbookFilename string,
+	playbookUUID string,
+	msgCounter int,
+	host *string,
+	tasks *results.AnsiblePlaybookJSONResultsPlayTaskItem,
+	hosts *results.AnsiblePlaybookJSONResultsPlayTaskHostsItem) *message.AnsibleRunnerJobEventYaml {
+	message := message.AnsibleRunnerJobEventYaml{
```
machacekondra, line 215: Maybe this could be resued for skipped/failed/on_ok, because the only difference is the Event name?

machacekondra (approved):

masayag (approved):

pkliczewski:
:tada:

PR #72, opened 2022-01-18
System metrics gathering

jakub-dzon:
This change set adds functionality of collecting system metrics from on-device node_exporter (http://localhost:9100/metrics) into local TSDB.


machacekondra (approved):

pkliczewski (approved):

internal/metrics/system.go
```diff
@@ -0,0 +1,62 @@
...
+	config       configuration.Manager
+}
+
+func NewSystemMetrics(daemon MetricsDaemon, config DeviceConfigurationProvider) *SystemMetrics {
+	expectedConfiguration := expectedConfiguration(config.GetDeviceConfiguration())
+
+	sm := &SystemMetrics{
+		daemon: daemon,
+	}
+	daemon.AddTarget("system", []string{NodeExporterMetricsEndpoint}, time.Duration(expectedConfiguration.Interval)*time.Second, nil)
```
eloycoto, line 35: I would move the "system" to a const, also added a if condition on metrics/workload if any the name is the same to log an error.

internal/metrics/system.go
```diff
@@ -0,0 +1,62 @@
...
+	newConfiguration := expectedConfiguration(*config.Configuration)
+	latestConfig := sm.latestConfig.Load()
+	if latestConfig != nil {
+		oldConfiguration := latestConfig.(*models.SystemMetricsConfiguration)
+		if oldConfiguration != nil && reflect.DeepEqual(newConfiguration, *oldConfiguration) {
+			return nil
+		}
+	}
+
+	sm.daemon.AddTarget("system", []string{NodeExporterMetricsEndpoint}, time.Duration(newConfiguration.Interval)*time.Second, nil)
```
eloycoto, line 50: We should add labels here, at least system one, if not can be mixed with another workload. Also, maybe the machine-id, if not multiple devices will have the same info, no?

internal/metrics/system_test.go
```diff
@@ -0,0 +1,203 @@
...
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/jakub-dzon/k4e-device-worker/internal/metrics"
+	"github.com/jakub-dzon/k4e-operator/models"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"time"
+)
+
+var _ = Describe("System", func() {
```
eloycoto, line 12: We should test double-update if add-workload is called again or not.

eloycoto (changes requested):

internal/metrics/system.go
jakub-dzon: The target name will be added as a label and if we don't allow using `system` workload target, it will be distinguishable.

The device ID label can be added when we send the metrics to the control plane - let's not waste memory on storing same label value multiple times.

internal/metrics/system_test.go
jakub-dzon: Done

internal/metrics/system.go
jakub-dzon: const - done, validation will be implemented separately in k4e-operator

internal/metrics/system.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/jakub-dzon/k4e-device-worker/internal/configuration"
+	"github.com/jakub-dzon/k4e-operator/models"
+	"reflect"
+	"sync/atomic"
+	"time"
+)
+
+const (
+	systemTargetName                     = "system"
+	DefaultSystemMetricsScrapingInterval = 60
```
ydayagi, line 13: we have a different interval for workloads and system? can u explain the rational?

ydayagi: we have a different interval for workloads and system? can u explain the rational?

internal/metrics/system.go
```diff
@@ -0,0 +1,63 @@
...
+	newConfiguration := expectedConfiguration(*config.Configuration)
+	latestConfig := sm.latestConfig.Load()
+	if latestConfig != nil {
+		oldConfiguration := latestConfig.(*models.SystemMetricsConfiguration)
+		if oldConfiguration != nil && reflect.DeepEqual(newConfiguration, *oldConfiguration) {
+			return nil
+		}
+	}
+
+	sm.daemon.AddTarget(systemTargetName, []string{NodeExporterMetricsEndpoint}, time.Duration(newConfiguration.Interval)*time.Second, nil)
```
ydayagi, line 51: Does every update add a target? don't we need to update an existing one or remove the existing one?

internal/metrics/system.go
```diff
@@ -0,0 +1,63 @@
...
+
+var defaultSystemMetricsConfiguration = models.SystemMetricsConfiguration{Interval: DefaultSystemMetricsScrapingInterval}
+
+//go:generate mockgen -package=metrics -destination=configuration_provider_mock.go . DeviceConfigurationProvider
+type DeviceConfigurationProvider interface {
+	GetDeviceConfiguration() models.DeviceConfiguration
+}
+
+type SystemMetrics struct {
+	latestConfig atomic.Value
```
ydayagi, line 25: Just a general question. I see that we use atomic instead of sync. Other locations in code use sync. Are we saying that both are OK for use in the project?

internal/metrics/system.go
jakub-dzon: Users might have different purpose metrics generated by their workloads and system metrics might be more or less important for them or be more or less variable, hence the possibility to set separate, custom interval. 

Default is the same for workload and system metrics and for convenience it is defined separately - to avoid accidental change to one when willing to modify only another.

internal/metrics/system.go
ydayagi: Just a thought I had. I think we could have just passed DeviceConfiguration to each feature and had a single update of the entire DeviceCnfiguration. This is instead of having each feature handle its own copy and sync.

internal/metrics/system.go
jakub-dzon: The underlying implementation of `AddTarget` always removes existing one first and adds the new one again.

internal/metrics/system.go
jakub-dzon: Yes, they are both OK and both have similar guarantees; I try to use atomic when possible, because it's less hassle than with mutexes.

That's an interesting idea with sharing the configuration. Sometimes the features will need to be informed that a change happened and we might end up with something similar - but maybe not in every case.

internal/metrics/system.go
ydayagi: the same can be said about specific metrics within each group. E.g. IP address metric is different from bytes-used metric. But both are system. But it doesn't hurt to have a separate interval.

internal/metrics/system.go
ydayagi: thanks. I guess I missed it.

ydayagi (approved):

eloycoto (approved):

PR #73, opened 2022-01-18
Evaluate collected hardware info

ydayagi:
Do not send unnecessary hadware information
Keep the API as is for now

https://issues.redhat.com/browse/ECOPROJECT-327

### DIFF
--- a/internal/hardware/hardware.go
+++ b/internal/hardware/hardware.go
@@ -1,28 +1,40 @@
 package hardware
 
 import (
-	"encoding/json"
 	"github.com/jakub-dzon/k4e-operator/models"
 	"github.com/openshift/assisted-installer-agent/src/inventory"
+	"github.com/openshift/assisted-installer-agent/src/util"
 )
 
 type Hardware struct {
 }
 
 func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
-	inv := inventory.ReadInventory(&inventory.Options{
-		GhwChrootRoot: "/",
-	})
+	d := util.NewDependencies("/")
+	cpu := inventory.GetCPU(d)
+	hostname := inventory.GetHostname(d)
+	systemVendor := inventory.GetVendor(d)
+	interfaces := inventory.GetInterfaces(d)
 
-	// Instead of copying field-by-field marshal to JSON and unmarshal to other struct
-	inventoryJson, err := json.Marshal(inv)
-	if err != nil {
-		return nil, err
-	}
 	hardwareInfo := models.HardwareInfo{}
-	err = json.Unmarshal(inventoryJson, &hardwareInfo)
-	if err != nil {
-		return nil, err
+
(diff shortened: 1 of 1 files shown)


jakub-dzon (commented):
Will it be possible to refactor the code in such a way that we don't even collect unnecessary hardware information? Instead of using `inventory.ReadInventory`, we could use underlying components directly (`inventory.GetCPU`, `inventory.GetHostname`, `inventory.GetVendor`, etc.) - it looks quite easy and could save some CPU cycles.

internal/hardware/hardware.go
```diff
@@ -14,16 +13,17 @@ func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
...
+
+	hardwareInfo.CPU = &models.CPU{
+		Architecture: inv.CPU.Architecture,
+		ModelName:    inv.CPU.ModelName,
+		Flags:        []string{},
 	}
 
+	hardwareInfo.Hostname = inv.Hostname
+
+	hardwareInfo.SystemVendor = (*models.SystemVendor)(inv.SystemVendor)
```
jakub-dzon, line 26: GPU info and flags details are missing (I understand they were removed in https://issues.redhat.com/browse/ECOPROJECT-327 for clarity and not as a requirement of removal) @masayag - is that correct?

ydayagi: I understood that it should be removed. But let's wait for @masayag cause he was in the meeting.

ydayagi:
> Will it be possible to refactor the code in such a way that we don't even collect unnecessary hardware information? Instead of using `inventory.ReadInventory`, we could use underlying components directly (`inventory.GetCPU`, `inventory.GetHostname`, `inventory.GetVendor`, etc.) - it looks quite easy and could save some CPU cycles.

sure. I'll try. if it works then great

internal/hardware/hardware.go
```diff
@@ -1,29 +1,29 @@
...
 }
 
 func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
-	inv := inventory.ReadInventory(&inventory.Options{
-		GhwChrootRoot: "/",
-	})
+	d := util.NewDependencies("/")
+	cpu := inventory.GetCPU(d)
+	hostname := inventory.GetHostname(d)
+	systemVendor := inventory.GetVendor(d)
```
masayag, line 16: as we learned, we'll have to collect and report IP addresses as well.

ydayagi: fixed

internal/hardware/hardware.go
ydayagi: @masayag

masayag: yes, they should be removed.

jakub-dzon (approved):

masayag (approved):

PR #74, opened 2022-01-19
Config: RegisterObserver init config.

eloycoto:
When yggd restart and any workload is already running, if the config
didn't change, it will not update the config to all registered observers.

With this commit, when an observer is created, has the option to send the
initial config to the Update() method, so the observers can run the initial
config and work against it.

### DIFF
--- a/internal/configuration/configuration.go
+++ b/internal/configuration/configuration.go
@@ -31,6 +31,7 @@ var (
 
 //go:generate mockgen -package=configuration -destination=configuration_mock.go . Observer
 type Observer interface {
+	Init(configuration models.DeviceConfigurationMessage) error
 	Update(configuration models.DeviceConfigurationMessage) error
 }
 
@@ -70,6 +71,12 @@ func NewConfigurationManager(dataDir string) *Manager {
 }
 
 func (m *Manager) RegisterObserver(observer Observer) {
+	// Always trigger the Init phase when register an observer to retrieve the
+	// current config.
+	err := observer.Init(*m.deviceConfiguration)
+	if err != nil {
+		log.Error("Running config init observer failed: ", err)
+	}
 	m.observers = append(m.observers, observer)
 }
 
--- a/internal/datatransfer/monitor.go
+++ b/internal/datatransfer/monitor.go
@@ -39,11 +39,6 @@ func NewMonitor(workloadsManager *workload.WorkloadManager, configManager *confi
 }
 
 func (m *Monitor) Start() {
-	// Made this to force the config load in case there is no storage defined yet
-	// via Update() method.
-	if err := m.ForceUpdate(); err != nil {
-		log.Errorf("cannot force-update datatransfer monitor. DeviceID: %s; err: %v", m.workloads.GetDeviceID(), err)
-	}
 	go func() {
 		for range m.ticker.C {
 			m.syncPaths()
@@ -240,11 +235,8 @@ func (m *Monitor) storeLastUpdateTime(workloadName string) {
 	m.lastSuccessfulSyncTimes[workloadName] = time.Now()
 }
 
-func (m *Monitor) ForceUpdate() error {
-	config := m.config.GetDeviceConfiguration()
-	return m.Update(models.DeviceConfigurationMessage{
-		Configuration: &config,
-	})
+func (m *Monitor) Init(configuration models.DeviceConfigurationMessage) error {
+	return m.Update(configuration)
 }
 
 func (m *Monitor) Update(configuration models.DeviceConfigurationMessage) error {
--- a/internal/metrics/system.go
+++ b/internal/metrics/system.go
@@ -1,11 +1,12 @@
 package metrics
 
 import (
-	"github.com/jakub-dzon/k4e-device-worker/internal/configuration"
-	"github.com/jakub-dzon/k4e-operator/models"
 	"reflect"
 	"sync/atomic"
 	"time"
+
+	"github.com/jakub-dzon/k4e-device-worker/internal/configuration"
+	"github.com/jakub-dzon/k4e-operator/models"
 )
 
 const (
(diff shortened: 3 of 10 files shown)


cmd/device-worker/main.go
```diff
@@ -100,32 +100,32 @@ func main() {
...
-	configManager.RegisterObserver(wl)
+	configManager.RegisterObserver(wl, true)
 	wl.RegisterObserver(workloadMetricWatcher)
 
 	hw := hardware2.Hardware{}
 
 	dataMonitor := datatransfer.NewMonitor(wl, configManager)
 	wl.RegisterObserver(dataMonitor)
-	configManager.RegisterObserver(dataMonitor)
+	configManager.RegisterObserver(dataMonitor, false)
```
eloycoto, line 120: @jakub-dzon  about this one, I'm not sure if we should wait until get the new s3 config, it's not clear to me.

ydayagi:
I think it is not so good for us to mix registration and update. The whole register mechanism becomes very complicated. My personal opinion is that observers are not needed. But I guess we're not going to change that. What I suggest (if possible) is to add a step after registering all observers that calls Update explicitly. Or add a step after registering the observers that calls observers' Update explicitly. Only for the observers that require it, of course.

internal/configuration/configuration.go
```diff
@@ -69,7 +69,13 @@ func NewConfigurationManager(dataDir string) *Manager {
 	return &mgr
 }
 
-func (m *Manager) RegisterObserver(observer Observer) {
+// RegisterObserver registers an Observer to the config, that will get called
+// when any config changes happens.
+// If withInitConfig is set, observer will be called with the current config.
+func (m *Manager) RegisterObserver(observer Observer, withInitConfig bool) {
```
jakub-dzon, line 75: I think managing dependencies between observers might be easier if the initial Update round would be performed at the same time for all of them, after they have been registered.
So the flow would be:
```
configManager.RegisterObserver(a, true)
configManager.RegisterObserver(b, false)
configManager.RegisterObserver(c, true)
// ...
configManager.Update() // or configManager.NotifyObservers() that would call Update on each observier with initialization enabled
```

cmd/device-worker/main.go
jakub-dzon: Right now it should be `false`, but we should adjust `Monitor.Start()` to not use `ForceUpdate` and set it to `true` then

internal/configuration/configuration.go
eloycoto: It'll not work, mainly because when workload start, emit some WorkloadStarted messages, and metrics does not have the config, etc.. It's a bit of chicken-egg issue.

cmd/device-worker/main.go
```diff
@@ -100,32 +100,32 @@ func main() {
 
 	// --- Client metrics configuration ---
 	metricsStore, err := metrics.NewTSDB(dataDir)
-	configManager.RegisterObserver(metricsStore)
+	configManager.RegisterObserver(metricsStore, true)
```
jakub-dzon, line 103: I will fix `TSDB` later not to be intialized with defaults in the factory method, but on the first update.

internal/configuration/configuration.go
jakub-dzon: Understood. :+1:

eloycoto:
> I think it is not so good for us to mix registration and update. The whole register mechanism becomes very complicated. My personal opinion is that observers are not needed. But I guess we're not going to change that. What I suggest (if possible) is to add a step after registering all observers that calls Update explicitly. Or add a step after registering the observers that calls observers' Update explicitly. Only for the observers that require it, of course.

@ydayagi  is this a hard request? could we defer to the next iteration after the migration on the project?

internal/configuration/configuration.go
```diff
@@ -69,7 +69,13 @@ func NewConfigurationManager(dataDir string) *Manager {
...
+	if withInitConfig {
```
masayag, line 76: I was thinking on a a slightly different way to implement this:
The `Observer` interface can be extended with `Init` method that will be responsible to making sure the observer is set properly (it will be each observer responsibility if to call Update from Init or perform any initialization logic or do nothing).
```
type Observer interface {
	Init(configuration models.DeviceConfigurationMessage) error
	Update(configuration models.DeviceConfigurationMessage) error
}
```
Then the main will remain mostly the same, without a need to specify for each observer how it should be started.
In other words, the responsibility to init the observer will be done within each observer.

eloycoto: Makes a lot of sense to me, I'll do in this way. Thanks @masayag

internal/metrics/system.go
```diff
@@ -39,6 +39,10 @@ func NewSystemMetrics(daemon MetricsDaemon, config DeviceConfigurationProvider)
 	return sm
 }
 
+func (sm *SystemMetrics) Init(config models.DeviceConfigurationMessage) error {
+	return sm.Update(config)
```
jakub-dzon, line 43: Initialization that is performed in `NewSystemMetrics` should be removed.

internal/datatransfer/monitor.go
```diff
@@ -41,9 +41,6 @@ func NewMonitor(workloadsManager *workload.WorkloadManager, configManager *confi
 func (m *Monitor) Start() {
 	// Made this to force the config load in case there is no storage defined yet
```
jakub-dzon, line 42: That comment should be removed

jakub-dzon (changes requested):

internal/datatransfer/monitor.go
eloycoto: :facepalm: sorry

pkliczewski (commented):
/lgtm

masayag (approved):

jakub-dzon (approved):

PR #75, opened 2022-01-19
System metrics allow list

jakub-dzon:
This PR adds support for system metrics allow-lists. 

By default only a limited set of system metrics will be collected and user can replace that default set by modifying their `EdgeDevice` CR.


ydayagi:
The allow list is for node exporter metrics, correct? Perhaps my question is too late, but can't we configure node exporter (or whatever is creating the system metrics) instead of changing the metrics daemon code?

jakub-dzon:
Yes, that's only for system metrics. 
We can configure a set of collectors (each produces tens of different metrics), but not set of metrics that we want to collect. Users might need only one specific metric from a collector

It is a good idea, to tune the collectors we use in our node_exporter. Now we don't know what customers might need and defaults seem reasonable.

ydayagi (approved):

masayag (approved):

pkliczewski (approved):

PR #76, opened 2022-01-21
System metrics disabling

jakub-dzon:
This PR introduces two changes:
 - disabling system metrics collection (including node_exporter shutdown) on  user request
 - initializing system metrics in new `Init` method instead of inside a factory method.


internal/workload/podman/podman.go
```diff
@@ -289,7 +289,7 @@ func (p *podman) GenerateSystemdService(podName string, monitoringInterval uint)
 		return nil, err
 	}
 
-	svc, err := service.NewSystemd(podName, report.Units)
+	svc, err := service.NewSystemd(podName, "pod-", report.Units)
```
pkliczewski, line 292: What about making it configurable?

jakub-dzon: The prefix? I think it is dictated by Podman. @machacekondra - is that right?

machacekondra: Correct:
```bash
$ podman generate systemd --help | grep pod-prefix
      --pod-prefix string         Systemd unit name prefix for pods (default "pod")

```

machacekondra (approved):

pkliczewski (approved):

eloycoto (approved):

PR #78, opened 2022-01-25
Adding os information and supporting graceful reboot

bardielle:


cmd/device-worker/main.go
```diff
@@ -150,6 +154,28 @@ func main() {
 	if err := s.Serve(l); err != nil {
```
jakub-dzon, line 154: I think that `s.Serve(l)` is blocking and the code after it will only be called when the process is being stopped. Is that the goal?

pkliczewski (commented):
What do you think about scheduling graceful reboot in the future? We could base it on reboot policy which would help to define maintenance window.

cmd/device-worker/main.go
bardielle: No, I need a blocking code for the channel 
what do you suggest me to do?

jakub-dzon: Would wrapping your code in a goroutine and running it before `s.Serve(l)` work correctly?

internal/configuration/configuration.go
```diff
@@ -97,7 +97,7 @@ func (m *Manager) GetSecrets() models.SecretList {
 }
 
 func (m *Manager) Update(message models.DeviceConfigurationMessage) error {
-
+	log.Infof("**** Update config was started")
```
jakub-dzon, line 100: Please remove the stars.

internal/server/deviceserver.go
```diff
@@ -27,15 +27,18 @@ func NewDeviceServer(configManager *configuration2.Manager, registrationManager
 // Send implements the "Send" method of the Worker gRPC service.
 func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error) {
 	go func() {
+		log.Infof("**** Get into SEND method")
```
jakub-dzon, line 30: Please remove the stars and in subsequent log messages.

jakub-dzon: Actually these should be either on Debug/Trace level or removed.

cmd/device-worker/main.go
```diff
@@ -147,9 +151,36 @@ func main() {
...
 
+	go ListenStartGracefulRebootChannel(err, wl, dataMonitor, hbs, deviceOs)
+
 	if err := s.Serve(l); err != nil {
 		log.Fatalf("cannot start worker server, err: %v", err)
 	}
+
+}
+
+func ListenStartGracefulRebootChannel(err error, wl *workload2.WorkloadManager, dataMonitor *datatransfer.Monitor, hbs *heartbeat2.Heartbeat, deviceOs *os2.OS) {
```
jakub-dzon, line 162: Please make it private.

cmd/device-worker/main.go
```diff
@@ -122,6 +123,11 @@ func main() {
 
 	hw := hardware2.Hardware{}
 
+	GracefulRebootChannel = make(chan string, 10)
```
jakub-dzon, line 126: Why is it 10?

cmd/device-worker/main.go
```diff
@@ -147,9 +151,36 @@ func main() {
...
+	for {
+		select {
+		case job := <-GracefulRebootChannel:
+			if job == os2.StartGracefulReboot {
+				log.Info("A graceful reboot request was received")
+				if err = wl.GracefulReboot(); err != nil {
+					log.Fatalf("cannot graceful reboot the workloads: %v", err)
+				}
+
+				if err = dataMonitor.Deregister(); err != nil {
```
jakub-dzon, line 173: I think it is worth deregistering `systemMetricsWatcher` and `metricsStore` - in that order.

jakub-dzon (changes requested):

cmd/device-worker/main.go
bardielle: I thought that we might need it for another things later but I will change it to be 1 (the default)

cmd/device-worker/main.go
bardielle: In the systemMetricsWatcher's Deregistration should be something more than deleting the target "system"?

cmd/device-worker/main.go
```diff
@@ -147,9 +151,45 @@ func main() {
...
 	}
+
+}
+
+func listenStartGracefulRebootChannel(err error, wl *workload2.WorkloadManager, dataMonitor *datatransfer.Monitor, systemMetricsWatcher *metrics.SystemMetrics, metricsStore *metrics.TSDB, hbs *heartbeat2.Heartbeat, deviceOs *os2.OS) {
+	// listen to the channel for getting StartGracefulReboot signal
+	for {
+		select {
+		case job := <-GracefulRebootChannel:
+			if job == os2.StartGracefulReboot {
```
masayag, line 167: is there an option for other values for this? else, we can just pick the type as of the channel as `struct{}{}`.

cmd/device-worker/main.go
```diff
@@ -29,6 +29,7 @@ import (
 )
 
 var yggdDispatchSocketAddr string
+var GracefulRebootChannel chan string
```
masayag, line 32: Can the scope of it can be reduced to the main function or at least reduce the visibility?
Also, the type implies there is importance for the value. It seems we use it as a "light' state-machine.
Perhaps it will be simpler to have two channels, one for initiating the shutdown and one for terminating the device.

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+
 const (
-	UnknownOsImageId = "unknown"
+	UnknownOsImageId                 = "unknown"
+	GreenbootHealthCheckFileName     = "/etc/greenboot/check/required.d/greenboot-health-check.sh"
+	GreenbooFailFileName             = "/etc/greenboot/red.d/bootfail.sh"
+	StartGracefulReboot              = "StartGracefulReboot"
+	FinishedGracefulReboot           = "FinishedGracefulReboot"
+	TimeoutToGracefulRebootInSeconds = 10
+	EdgeConfFileName                 = "/etc/ostree/remotes.d/edge.conf"
```
masayag, line 23: Were you able to verify this is common for all arch/OS?
also, since used internally, we can reduce the visibility to be unexported.

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+	return o.Update(config)
+}
+
+func (o *OS) Update(configuration models.DeviceConfigurationMessage) error {
+	newOSInfo := configuration.Configuration.Os
+	if newOSInfo == nil {
+		return fmt.Errorf("cannot retrieve configuration info.")
+	}
+
+	if newOSInfo.HostedImagesURL != o.HostedImagesURL {
```
masayag, line 109: What should happen if automaticUpgrade is set to false? or to true? 
Shouldn't the `AutomaticUpdatePolicy` be updated accordingly not to/to trigger auto update by rpm-ostreed-automatic.timer? 
https://www.mankier.com/8/rpm-ostreed-automatic.service
https://www.mankier.com/5/rpm-ostreed.conf

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+		stdout, err := cmd.Output()
+
+		if err != nil {
+			log.Error("Failed to run 'rpm-ostree update --preview'")
+			return err
+		}
+
+		// TODO check - if there is a multiple commits?
+		if !strings.Contains(string(stdout), newOSInfo.CommitID){
+			log.Error("Failed to run 'rpm-ostree update --preview'")
```
masayag, line 132: that's not the right error message.

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+	}
+
+	if newOSInfo.CommitID != o.RequestedOsCommit {
+		log.Infof("The requested commit ID has been changed to %s", newOSInfo.CommitID)
+		o.RequestedOsCommit = newOSInfo.CommitID
+		cmd := exec.Command("rpm-ostree", "update", "--preview")
+		stdout, err := cmd.Output()
+
+		if err != nil {
+			log.Error("Failed to run 'rpm-ostree update --preview'")
```
masayag, line 126: will it be useful to log the error itself?

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+			_, err = greenbootHealthCheckFile.Write([]byte(GreenbootHealthCheckScript))
+			if err != nil {
+				return err
+			}
+			greenbootHealthCheckFile.Chmod(755)
+		} else {
+			return err
+		}
+	}
+
```
masayag, line 235: pls remove extra space line

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+			}
+			_, err = greenbootHealthCheckFile.Write([]byte(GreenbootHealthCheckScript))
+			if err != nil {
+				return err
+			}
+			greenbootHealthCheckFile.Chmod(755)
+		} else {
+			return err
+		}
+	}
```
masayag, line 234: seems this can be generified and extracted into a method, e.g.: `ensureScriptExists(name, content) err` and reused for both files.

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+			return fmt.Errorf("cannot find the new commit hash. %s", newOSInfo.CommitID)
+		}
+
+		err = updateGreenbootScripts()
+		if err != nil {
+			log.Error("Failed to update greenboot scripts")
+			return err
+		}
+
+		cmd = exec.Command("rpm-ostree", "upgrade")
```
masayag, line 142: This will upgrade to the latest commit available, right? How does it being verified to be the exact commit as requested?

internal/os/os.go
```diff
@@ -1,12 +1,260 @@
...
+			return err
+		}
+
+		o.GracefulRebootFlow()
+
+		cmd = exec.Command("systemctl", "reboot")
+		stdout, err = cmd.Output()
+
+		if err != nil {
+			return fmt.Errorf("failed to run 'rpm-ostree reboot': %s", err)
```
masayag, line 156: pls update message s/rpm-ostree reboot/systemctl reboot

masayag (changes requested):

cmd/device-worker/main.go
jakub-dzon: No, that's it. 
It needs to be done before `metricsStore` or else there might be errors (from writing to a closed DB) in the logs.

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION = 1.0
+VERSION = 1.2.9
```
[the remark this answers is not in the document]
bardielle: Im running it meanwhile so I changed the version - I will return it to be 1.0

internal/workload/manager.go
```diff
@@ -283,6 +283,19 @@ func (w *WorkloadManager) RegisterObserver(observer Observer) {
 	w.workloads.RegisterObserver(observer)
 }
 
+func (w *WorkloadManager) GracefulReboot() error{
```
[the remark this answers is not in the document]
bardielle: :+1:  I changed it to StopWorkloads

internal/metrics/system.go
```diff
@@ -69,6 +69,12 @@ func (sm *SystemMetrics) Update(config models.DeviceConfigurationMessage) error
 	return nil
 }
 
+func (sm *SystemMetrics) Deregister() error {
```
jakub-dzon, line 72: That method is already in `main` (https://github.com/jakub-dzon/k4e-device-worker/blob/7ff785b3ed54ec8f53173b982cc3851ad0469ae8/internal/metrics/system.go#L73). 
Please make sure that you work on the latest changes.

internal/os/os.go
bardielle: In the previous check

bardielle: I can use rpm-ostree deploy + commitID 
but it might cost a lot of traffic if we have a lot of devices so I preferred to check by using the "preview" cmd

jakub-dzon (approved):

jakub-dzon (changes requested):
Approved too early.

internal/os/os.go
bardielle: updated it

jakub-dzon (approved):

internal/os/os.go
masayag: @bardielle any update on this?

internal/workload/podman/podman.go
```diff
@@ -63,9 +64,11 @@ type PodReport struct {
 func (p *PodReport) AppendContainer(c *ContainerReport) {
 	p.Containers = append(p.Containers, c)
 }
+const DefaultTimeoutForStopping int = 5
```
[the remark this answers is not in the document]
masayag: @bardielle pls add units to constant name

cmd/device-worker/main.go
```diff
@@ -147,9 +152,46 @@ func main() {
...
+
+			if err := metricsStore.Deregister(); err != nil {
+				log.Fatalf("cannot graceful reboot metricsStore: %v", err)
+			}
+
+			if err := hbs.Deregister(); err != nil {
+				log.Fatalf("cannot graceful reboot the heartbeat service: %v", err)
+			}
+
+			gracefulRebootChannel <- struct{}{}
```
masayag, line 192: can't this lead to another invocation of this for loop? 
shouldn't we break outside or use a dedicated channel for graceful shutdown completion?

jakub-dzon: I think Moti is right that there might be some unexpected behavior as a result of that call. I suspect there might be a race condition and in some cases that may result in an infinite loop.

bardielle: I tested it in a playground env 
but for making it clear I will add

internal/workload/podman/podman.go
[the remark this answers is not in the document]
bardielle: I was looking for a documentation in podman but didn't find about it... still searching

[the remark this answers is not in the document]
bardielle: https://docs.podman.io/en/latest/markdown/podman-stop.1.html

internal/os/os.go
bardielle: I found that there is an option for configuring the directory creation by a flag add-remotes-config-dir : https://manpages.debian.org/testing/ostree/ostree.repo-config.5.en.html

but not about the  file name

bardielle: It seems to be set it in the ks file - ```ostreesetup --nogpg --url=http://192.168.122.233/device-worker/repo --osname=rhel --remote=edge --ref=rhel/8/x86_64/edge```

bardielle: After investigating it seems the ostree looks for the remotes in the content of the file so the name is not relevant but the remote key name is the thing that importing to mention (which is edge in our case but might be change by the user)

I opened ticket for that

masayag (approved):

jakub-dzon (approved):

internal/os/greenboot_scripts.go
```diff
@@ -12,40 +12,36 @@ echo "" >> /var/roothome/greenboot.log`
...
+  exit 1
 fi
 
-if [ "$(systemctl is-active podman.socket)" = "active" ]; then
-echo "podman.socket is active, check passed!"
-else
-echo "podman.socket is not active, check failed!"
-exit 1
-fi
+CheckServiceIsActive(){
```
masayag, line 24: just in term of naming convention, pls change to check_is_service_active
also, for simplifying, instead of passing a similar text as arguments, just the name of the service is enough.
so replace a call such as:
```
CheckServiceIsActive "$(systemctl is-active yggdrasild.service)" "yggdrasild.service is active, check passed!" "yggdrasild.service is not active, retrying!" "yggdrasild"
```
with
```
check_is_service_active yggdrasild.service
check_is_service_active nftables.service
...
```
and  both the messages and the command to invoke will be created inside the function, since they are sharing the same format and logic.

masayag (approved):

masayag (approved):

cmd/device-worker/main.go
```diff
@@ -122,6 +123,11 @@ func main() {
 
 	hw := hardware2.Hardware{}
 
+	gracefulRebootChannel = make(chan struct{})
+	deviceOs := os2.NewOS(gracefulRebootChannel)
+	configManager.RegisterObserver(deviceOs)
+
```
masayag, line 129: nit there is an extra empty line

PR #79, opened 2022-01-25
Change systemd unit names to pod IDs

machacekondra:
This PR changes the systemd unit names to use pod IDs, because for some
use cases the pod names may not work.

For example when container name and pod name has same for different
podman version the podman generate different pod name.

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -3,8 +3,9 @@ package podman
 import (
 	"context"
 	"fmt"
-	"github.com/jakub-dzon/k4e-device-worker/internal/service"
 	"strings"
+
+	"github.com/jakub-dzon/k4e-device-worker/internal/service"
 
 	"git.sr.ht/~spc/go-log"
 	podmanEvents "github.com/containers/podman/v3/libpod/events"
@@ -282,14 +283,13 @@ func (p *podman) UpdateSecret(name, data string) error {
 	return p.CreateSecret(name, data)
 }
 
-func (p *podman) GenerateSystemdService(podName string, monitoringInterval uint) (service.Service, error) {
-	useName := true
-	report, err := generate.Systemd(p.podmanConnection, podName, &generate.SystemdOptions{UseName: &useName, RestartSec: &monitoringInterval})
+func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (service.Service, error) {
+	report, err := generate.Systemd(p.podmanConnection, podId, &generate.SystemdOptions{RestartSec: &monitoringInterval})
 	if err != nil {
 		return nil, err
 	}
 
-	svc, err := service.NewSystemd(podName, "pod-", report.Units)
+	svc, err := service.NewSystemd(podId, "pod-", report.Units)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/workload/wrapper.go
+++ b/internal/workload/wrapper.go
@@ -2,6 +2,7 @@ package workload
 
 import (
 	"fmt"
+
 	"github.com/jakub-dzon/k4e-device-worker/internal/service"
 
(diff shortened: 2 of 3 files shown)


machacekondra:
cc @jakub-dzon

internal/service/systemd.go
```diff
@@ -164,7 +163,7 @@ func (s *systemd) Disable() error {
 }
 
 func (s *systemd) serviceName(serviceName string) string {
-	return s.servicePrefix + serviceName + ServiceSuffix
+	return ServicePrefix + serviceName + ServiceSuffix
```
jakub-dzon, line 166: That breaks `node_exporter` service control - it requires blank prefix.

jakub-dzon (changes requested):

internal/workload/podman/podman.go
```diff
@@ -282,14 +283,15 @@ func (p *podman) UpdateSecret(name, data string) error {
 	return p.CreateSecret(name, data)
 }
 
-func (p *podman) GenerateSystemdService(podName string, monitoringInterval uint) (service.Service, error) {
-	useName := true
-	report, err := generate.Systemd(p.podmanConnection, podName, &generate.SystemdOptions{UseName: &useName, RestartSec: &monitoringInterval})
+func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (service.Service, error) {
+	//report, err := generate.Systemd(p.podmanConnection, podId, &generate.SystemdOptions{RestartSec: &monitoringInterval})
```
jakub-dzon, line 287: Please remove commented-out code.

internal/workload/podman/podman.go
```diff
@@ -282,14 +283,15 @@ func (p *podman) UpdateSecret(name, data string) error {
...
+	log.Errorf("Passing the podId: %v", podId)
```
jakub-dzon, line 288: Please change logging level or remove.

jakub-dzon (changes requested):

internal/workload/podman/podman.go
machacekondra: Ah, thanks. I've found bug in podman, that if you pass empty podId, it crashes, so that's why I added those here and I forgot to remove.

internal/service/systemd.go
machacekondra: Done.

internal/workload/podman/podman.go
```diff
@@ -282,14 +283,13 @@ func (p *podman) UpdateSecret(name, data string) error {
...
-	useName := true
-	report, err := generate.Systemd(p.podmanConnection, podName, &generate.SystemdOptions{UseName: &useName, RestartSec: &monitoringInterval})
+func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (service.Service, error) {
+	report, err := generate.Systemd(p.podmanConnection, podId, &generate.SystemdOptions{RestartSec: &monitoringInterval})
 	if err != nil {
 		return nil, err
 	}
 
-	svc, err := service.NewSystemd(podName, "pod-", report.Units)
+	svc, err := service.NewSystemd(podId, "pod-", report.Units)
```
jakub-dzon, line 292: From what I understand, `podId` will become part of the service name (`pod-<podId>.service`), which will become pretty lengthy; won't that be a problem for systemd? I don't thing it would matter a lot for debuggability.

machacekondra: I've tested and it looks like:
```bash
systemctl status pod-ee7b06ed00c24e66528121a2cd7c1d721e7897a99ad25145d63c4cc529427e0b.service
```
so yes it's lenghty, but works.

jakub-dzon (approved):

eloycoto (approved):

PR #82, opened 2022-01-26
Fix make dist

jordigilh:
Fixes 2 issues in the Makefile:
* Failure to copy the .tar.gz to the `rpmbuild/SOURCES` directory as the directory does not exist at that point. Using the `install -D` command solves the issue.
  * How to reproduce:
    * Delete the ~/rpmbuild` directory
    * Run `make rpm-build`
    * Expected outcome: Successful build
    * Actual result: Failed to copy `.tar.gz` to `~/rpmbuild/SOURCES/`

* `make dist` only creates the rpms for a single architecture (when run the second time it will complete the remaining one). I suspect it has to do with the fact that both `rpm` and `rpm-arm64` depend on the same target `rpm-build` and `make` skips calling the `rpm-build` the second time as it has been already completed previously.
  *  How To reproduce:
     *  Delete the `~/rpmbuild` directory
     * Run `make dist`
     * Check that both aarch64 and x86_64 directories exist in `~/rpmbuild/RPMS`:
     * Expected result: 2 directories with an rpm in each directory.
     * Actual result: only the x86_64 directory and rpm exist

@eloycoto @jakub-dzon please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ generate: ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject
 	$(Q) go generate ./...
 
 .PHONY: vendor
+
 vendor:
 	go mod tidy
 	go mod vendor
@@ -100,7 +101,7 @@ rpm-tarball:
 	    | gzip > k4e-agent-$(VERSION).tar.gz
 
 rpm-src: rpm-tarball
-	cp k4e-agent-$(VERSION).tar.gz $(HOME)/rpmbuild/SOURCES/
+	install -D -m644 k4e-agent-$(VERSION).tar.gz --target-directory `rpmbuild -E %_sourcedir`
 	rpmbuild -bs \
 		-D "VERSION $(VERSION)" \
 		-D "RELEASE $(RELEASE)" \
@@ -114,11 +115,10 @@ rpm-build: rpm-src
 	rpmbuild $(RPMBUILD_OPTS) --rebuild $(HOME)/rpmbuild/SRPMS/k4e-agent-$(VERSION)-$(RELEASE).*.src.rpm
 
 rpm: ## Create rpm build
-rpm: rpm-build
+	RPMBUILD_OPTS=--target=x86_64 $(MAKE) rpm-build
 
 rpm-arm64: ## Create rpm build for arm64
-rpm-arm64: RPMBUILD_OPTS=--target=aarch64
-rpm-arm64: rpm-build
+	RPMBUILD_OPTS=--target=aarch64  $(MAKE) rpm-build
 
 dist: ## Create distribution packages
 dist: build build-arm64 rpm rpm-arm64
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -100,7 +101,7 @@ rpm-tarball:
 	    | gzip > k4e-agent-$(VERSION).tar.gz
 
 rpm-src: rpm-tarball
-	cp k4e-agent-$(VERSION).tar.gz $(HOME)/rpmbuild/SOURCES/
+	install -D -m644 k4e-agent-$(VERSION).tar.gz --target-directory `rpmbuild -E %_sourcedir`
```
eloycoto, line 104: I'm ok with that, if you run make rpm-copr, this will work?

here is how it works:
https://github.com/jakub-dzon/k4e-device-worker/blob/main/.github/workflows/rpm.yaml

jordigilh: Yep, it builds the rpm files and this [line](https://github.com/jakub-dzon/k4e-device-worker/blob/main/.github/workflows/rpm.yaml#L30) is no longer necessary.


```bash
[root@83a558d5e6fa k4e-device-worker]# rm -Rf ~/rpmbuild/
[root@83a558d5e6fa k4e-device-worker]# make rpm-tarball
which: no ginkgo in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
which: no gover in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
(git archive --prefix k4e-agent-1.0/ HEAD ) \
    | gzip > k4e-agent-1.0.tar.gz
[root@83a558d5e6fa k4e-device-worker]# make rpm-src
which: no ginkgo in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
which: no gover in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
(git archive --prefix k4e-agent-1.0/ HEAD ) \
    | gzip > k4e-agent-1.0.tar.gz
install -D -m644 k4e-agent-1.0.tar.gz --target-directory `rpmbuild -E %_sourcedir`
rpmbuild -bs \
	-D "VERSION 1.0" \
	-D "RELEASE 1" \
	-D "_libexecdir /usr/libexec" \
	--buildroot /root/k4e-device-worker/dist ./k4e-agent.spec
Wrote: /root/rpmbuild/SRPMS/k4e-agent-1.0-1.el8.src.rpm
```

I did not succeed in running the last command because I could not find the `copr-cli` package in CentOS. But I wager that it does not matter at that point since the `.src.rpm` file is already in place.

eloycoto: Then works for me :-) 

Thanks jordi

eloycoto (approved):

jakub-dzon (approved):

PR #84, opened 2022-02-02
Metrics: Added allowlist filtering

eloycoto:
With this commit, users can define what metrics want to to retrieve
on workload level.

Related-to: https://github.com/project-flotta/flotta-operator/commit/5b67dd38215b6a110a6293c9ac35875e9417daa0
Fix: ECOPROJECT-495

### DIFF
--- a/internal/metrics/workload.go
+++ b/internal/metrics/workload.go
@@ -68,6 +68,11 @@ func (wrkM *WorkloadMetrics) WorkloadStarted(workloadName string, report []*podm
 			continue
 		}
 
+		var filter SampleFilter = &PermissiveAllowList{}
+		if cfg.Metrics.AllowList != nil {
+			filter = NewRestrictiveAllowList(cfg.Metrics.AllowList)
+		}
+
 		urls := []string{}
 		for _, workloadReport := range report {
 			urls = append(urls, getWorkloadUrls(workloadReport, cfg)...)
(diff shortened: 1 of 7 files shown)


internal/metrics/workload.go
```diff
@@ -68,6 +68,11 @@ func (wrkM *WorkloadMetrics) WorkloadStarted(workloadName string, report []*podm
 			continue
 		}
 
+		var filter SampleFilter = &PermissiveAllowList{}
+		if cfg.Metrics.AllowList != nil && len(cfg.Metrics.AllowList.Names) > 0 {
```
eloycoto, line 72: @jakub-dzon  not sure here, if there are no Names defined should we do Restrictive or Permissive.

jakub-dzon: IMO when it's nil, it should be Permissive, if empty - Restrictive.

internal/metrics/workload.go
```diff
@@ -78,7 +83,11 @@ func (wrkM *WorkloadMetrics) WorkloadStarted(workloadName string, report []*podm
 			interval = cfg.Metrics.Interval
 		}
 		// log for this is part of the AddTarget function
-		wrkM.daemon.AddTarget(workload.Name, urls, time.Duration(interval)*time.Second)
+		wrkM.daemon.AddFilteredTarget(
```
jakub-dzon, line 86: Can we remove `AddTarget` at this point?

internal/metrics/workload.go
eloycoto: :+1:  I was not sure about.

internal/metrics/workload.go
eloycoto: Removed the AddFilteredTarget and rename to AddTarget.

internal/metrics/workload.go
```diff
@@ -68,6 +68,11 @@ func (wrkM *WorkloadMetrics) WorkloadStarted(workloadName string, report []*podm
 			continue
 		}
 
+		var filter SampleFilter = &PermissiveAllowList{}
```
jakub-dzon, line 71: `filter SampleFilter := &PermissiveAllowList{}`

jakub-dzon (approved):

masayag (approved):

PR #91, opened 2022-02-11
Add test image build after PR merge

machacekondra:
This commit add a new workflow job to build the container image for the
e2e tests.

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -1,0 +1,22 @@
+---
+name: Container image build
+on:
+  push:
+    branches: [ main ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out proper version of sources
+        uses: actions/checkout@v1
+
+      - name: Login to quay
+        env:
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          echo "$QUAY_TOKEN" | docker login quay.io --username project-flotta+github --password-stdin
+
+      - name: Build & push the container image to quay
+        run: |
+          make deploy-container-image
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,0 +1,35 @@
+FROM quay.io/podman/stable:v3.4.4
+
+WORKDIR /project
+
+# Setup Yggdrasil & device-worker repo:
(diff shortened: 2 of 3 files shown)


.github/workflows/container.yaml
```diff
@@ -0,0 +1,23 @@
+---
+name: Container image build
+on:
+  pull_request:
+    types: [ closed ]
```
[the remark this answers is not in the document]
machacekondra: If I am not mistaken if we will use: `on_push: master` then this job will be executed also for the opened PR, which I didn't want to. I thought it would be better to build&push the docker image only for merged PRs.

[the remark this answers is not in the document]
eloycoto: With this setup, only when push to main branch, and the branch will be protected in this repo as the operator.  

```
on:
  push:
    branches: [ main ]
```

tools/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM quay.io/podman/stable
```
eloycoto, line 1: Should we pin the version here?

tools/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+WORKDIR /project
+
+# Setup Yggdrasil & device-worker repo:
+RUN curl -k https://copr.fedorainfracloud.org/coprs/ondramachacek/yggdrasil/repo/fedora-34/ondramachacek-yggdrasil-fedora-34.repo -o /etc/yum.repos.d/yggd.repo
+RUN curl -k https://copr.fedorainfracloud.org/coprs/eloyocoto/k4e-test/repo/fedora-34/eloyocoto-k4e-test-fedora-34.repo -o /etc/yum.repos.d/flotta-agent.repo && \
+    sed -i 's/\$releasever/34/g' /etc/yum.repos.d/flotta-agent.repo && \
+    sed -i 's/\$basearch/x86_64/g' /etc/yum.repos.d/flotta-agent.repo
+
+# Yum dependencies
+RUN yum install -y golang openssl procps-ng dmidecode flotta-agent nc
```
eloycoto, line 12: Why golang is needed here?

machacekondra: Ah, right. It's leftover because I've previously used the ygg from source not from RPM. Thanks!

tools/Dockerfile
machacekondra: Yes!

eloycoto (approved):
Thanks, awesome work.

pkliczewski (approved):

PR #95, opened 2022-02-17
WIP: Podman: Added Logs method

eloycoto:
Added a new Log method to Podman interface, so can retrieve log to the
given writer interface.

This is needed for the log collection, so pod logs can be send to
another transport in the near future.

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -3,6 +3,7 @@ package podman
 import (
 	"context"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/project-flotta/flotta-device-worker/internal/service"
@@ -29,6 +30,11 @@ const (
 	podmanStop   = string(podmanEvents.Stop)
 )
 
+var (
+	logTailLines = "1"
+	boolTrue     = true
+)
+
 //go:generate mockgen -package=podman -destination=mock_podman.go . Podman
 type Podman interface {
 	List() ([]api2.WorkloadInfo, error)
@@ -42,6 +48,7 @@ type Podman interface {
 	UpdateSecret(name, data string) error
 	Exists(workloadId string) (bool, error)
 	GenerateSystemdService(podName string, monitoringInterval uint) (service.Service, error)
+	Logs(podID string, res io.Writer) (context.CancelFunc, error)
 }
 
 type PodmanEvent struct {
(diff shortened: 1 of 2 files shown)


internal/workload/podman/podman.go
```diff
@@ -316,3 +323,56 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+			if err != nil {
+				log.Errorf("cannot get container '%v' logs: %v", name, err)
+			}
+		}(container.ID)
+	}
+
+	go func() {
+		for {
+			select {
+			case line := <-stdoutCh:
```
pkliczewski, line 358: why not to use multi value case? Lines 359-362 and 364-267 are identical.

eloycoto: select does not have multiple value cases no?

pkliczewski: right, only switches have multi value cases

pkliczewski (approved):

internal/workload/podman/podman.go
```diff
@@ -316,3 +323,56 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+		Stderr: &boolTrue,
+		Stdout: &boolTrue,
+		Tail:   &logTailLines,
+	}
+
+	ctx, cancel := context.WithCancel(p.podmanConnection)
+	stdoutCh := make(chan string, 100)
+	stderrCh := make(chan string, 100)
+
+	for _, container := range podInfo.Containers {
```
jakub-dzon, line 346: This loop starts logs streaming for all containers using one set of channels and as a result the log lines from multiple containers will be mixed in one writer (possibly even not monotonically). 

Do I understand the code correctly? If so, I don't think that is desired.

eloycoto: yes, the logs are per pod, how many use cases we're going to have of multiple containers in each edge device? should we add more logic in there per container, etc..? I do not think at this stage. We can return an array of io.Reader, but I'm not sure if it's going to be helpful at all. 

Saying that, Podman added a pause container too, but as I said, multiple readers can be returned.

jakub-dzon: I think we should handle multi-container pods correctly from the start. It could be returning multiple readers or annotating each log line with container ID. I would prefer feeding the logs into separate structures that could easily be transformed into files.

jakub-dzon:
How do you plan to use the log retrieval functionality to guarantee that users eventually get all the log lines and the logs are not transferred in full each time?

eloycoto:
> How do you plan to use the log retrieval functionality to guarantee that users eventually get all the log lines and the logs are not transferred in full each time?


Made a RFC doc for that: https://docs.google.com/document/d/1_hWAmLxZWkivIPe_XcoR6xxomuHFTqw4aB_A30MYOUc/edit#heading=h.sflsgfnz0ut7

internal/workload/podman/podman.go
eloycoto: Ok

eloycoto:
Change is so big with the containerName etc.. that I prefer to close this one and open a  new one with the valid context, commits, pr description.

PR #96, opened 2022-02-21
Podman: Added Logs method

eloycoto:
Added a new Log method to Podman interface, so can retrieve log to the
given writer interface with the container ID.

There is new channel per each container, per each pod, and two loops
that read the same Reader, but to have a prefix per each container, we
need to have in this way.

This is needed for the log collection, so pod logs can be send to
another transport in the near future.

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -1,8 +1,11 @@
 package podman
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/project-flotta/flotta-device-worker/internal/service"
@@ -29,6 +32,11 @@ const (
 	podmanStop   = string(podmanEvents.Stop)
 )
 
+var (
+	logTailLines = "1"
+	boolTrue     = true
+)
+
 //go:generate mockgen -package=podman -destination=mock_podman.go . Podman
 type Podman interface {
 	List() ([]api2.WorkloadInfo, error)
@@ -42,6 +50,7 @@ type Podman interface {
 	UpdateSecret(name, data string) error
 	Exists(workloadId string) (bool, error)
 	GenerateSystemdService(podName string, monitoringInterval uint) (service.Service, error)
+	Logs(podID string, res io.Writer) (context.CancelFunc, error)
 }
 
 type PodmanEvent struct {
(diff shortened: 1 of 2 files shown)


internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+					}
+
+					if !isPrefix {
+						resultLine = append(resultLine, line...)
+					}
+				}
+				if len(resultLine) > 0 {
+					_, err := res.Write([]byte(fmt.Sprintf("%s: %s\n", containerID, resultLine)))
+					if err != nil {
+						log.Errorf("Cannot write container log line: %v", err)
```
masayag, line 367: WDYT of including the pod-id/container id in the message?

internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+// Retrieve all pods logs and send that to the given io.Writer
+func (p *podman) Logs(podID string, res io.Writer) (context.CancelFunc, error) {
+
+	podInfo, err := pods.Inspect(p.podmanConnection, podID, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithCancel(p.podmanConnection)
+	writers := map[string]*bufio.Reader{}
```
masayag, line 338: it is named `writers` but it is actually a map of readers, no?

internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+	for _, container := range podInfo.Containers {
+		res := &bytes.Buffer{}
```
masayag, line 340: isn't there a risk of consuming too much memory for it? shouldn't it be dumped to a file?

internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+		Stdout: &boolTrue,
+		Tail:   &logTailLines,
+	}
+	stdoutCh := make(chan string, 100)
+	stderrCh := make(chan string, 100)
+
+	go func() {
+		err := containers.Logs(ctx, containerID, &opts, stdoutCh, stderrCh)
+		if err != nil {
+			log.Error("cannot get logs for container", err)
```
masayag, line 395: pls include pod-id/container-id in message

internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+}
+
+func (p *podman) containerLog(ctx context.Context, containerID string, res io.Writer) error {
+	opts := containers.LogOptions{
+		Follow: &boolTrue,
+		Stderr: &boolTrue,
+		Stdout: &boolTrue,
+		Tail:   &logTailLines,
+	}
+	stdoutCh := make(chan string, 100)
```
masayag, line 389: why buffered channel?  any why 100 if this is per-container?

internal/workload/podman/podman.go
eloycoto: no, no a risk, another process will eat all and have an allowed buffer for it.

internal/workload/podman/podman.go
eloycoto: wanted to have by pod, but we need to have per container in case one pod has multiple containers. About buffer, to not block anything if pod logs are faster than pod reads.

masayag (approved):

internal/workload/podman/podman.go
```diff
@@ -316,3 +325,96 @@ func (p *podman) GenerateSystemdService(podId string, monitoringInterval uint) (
...
+	stderrCh := make(chan string, 100)
+
+	go func() {
+		err := containers.Logs(ctx, containerID, &opts, stdoutCh, stderrCh)
+		if err != nil {
+			log.Errorf("cannot get logs for container '%s': %s", containerID, err)
```
masayag, line 395: the second arg should be %v for err

pkliczewski (approved):

jakub-dzon (commented):
LGTM

PR #98, opened 2022-02-21
Update rpm source

masayag:

### DIFF
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -7,7 +7,7 @@ Summary:    Agent application for the Flotta Edge Management solution
 ExclusiveArch: %{go_arches}
 Group:      Flotta
 License:    GPL
-Source:    %{name}-%{version}.tar.gz
+Source0:    %{name}-%{version}.tar.gz
 
 BuildRequires:  golang
 
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -7,7 +7,7 @@ Summary:    Agent application for the Flotta Edge Management solution
 ExclusiveArch: %{go_arches}
 Group:      Flotta
 License:    GPL
-Source0:    https://github.com/jakub-dzon/%{name}-%{version}.tar.gz
+Source:    %{name}-%{version}.tar.gz
 
 BuildRequires:  golang
 


jakub-dzon (approved):

flotta-agent.spec
```diff
@@ -7,7 +7,7 @@ Summary:    Agent application for the Flotta Edge Management solution
 ExclusiveArch: %{go_arches}
 Group:      Flotta
 License:    GPL
-Source0:    https://github.com/jakub-dzon/%{name}-%{version}.tar.gz
+Source0:    https://github.com/project-flotta/%{name}-%{version}.tar.gz
```
eloycoto, line 10: this will not work at all no? Should be flotta-device-worker and there is no target version. Should be `%{name}-%{version}.tar.gz`

eloycoto (changes requested):

flotta-agent.spec
masayag: you're correct. however, do we push it to github? should `Source0` be removed or are we missing some functionality?

eloycoto: if you have like that, it'll fail when you do spectool.. so I wouldn't use the URL at all.

flotta-agent.spec
```diff
@@ -7,7 +7,6 @@ Summary:    Agent application for the Flotta Edge Management solution
 ExclusiveArch: %{go_arches}
 Group:      Flotta
 License:    GPL
-Source0:    https://github.com/jakub-dzon/%{name}-%{version}.tar.gz
```
eloycoto, line 10: If you remove Source0, the line 29 will fail: 
```
tar fx %{SOURCE0}
```
As I said, sources should be: `%{name}-%{version}.tar.gz`

eloycoto (changes requested):

flotta-agent.spec
masayag: I get it now. I missed the fact that we shouldn't refer to github at all

masayag: looks good now:
```
→ sudo rpm -qi  /root/rpmbuild/RPMS/x86_64/flotta-agent-1.0-1.fc33.x86_64.rpm
Name        : flotta-agent
Version     : 1.0
Release     : 1.fc33
Architecture: x86_64
Install Date: (not installed)
Group       : Flotta
Size        : 41178977
License     : GPL
Signature   : (none)
Source RPM  : flotta-agent-1.0-1.fc33.src.rpm
Build Date  : Tue 22 Feb 2022 02:45:20 PM IST
Build Host  : localhost
Summary     : Agent application for the Flotta Edge Management solution
Description :
The Flotta agent communicates with the Flotta control plane. It reports the status of the appliance and of the running PODs/containers. Agent is responsible for starting and stopping PODs that are based on commands from the control plane.
```

eloycoto (approved):

pkliczewski (approved):

PR #99, opened 2022-02-25
Don't start yggdrasil in test docker image

machacekondra:


jakub-dzon:
Shouldn't flotta-agent-1.0.tar.gz file be removed from the PR?

masayag:
@machacekondra could you share the reason for it?
Is the assumption that setting the machine-id and the starting the yggd should be done only when it is actually being started on the machine (not to get machine-id of build machine) and start the server only when the identify of the cluster is known not to fail on registration?

machacekondra:
@eloycoto can share more I think. But now we don't fetch the machine-id and set it as client-id in case client-id is not passed. The only way is to use CN from certificate or use client-id in config or command line, when starting yggdrasil. That's the reason we can't set it during image build, but we must set it after the container is started, and so we can't start yggdrasil service, and we must start it after we confgiure the client id, which we must set in runtime for the test, so we can support multiple edgedevices.

pkliczewski (approved):

PR #101, opened 2022-03-01
Collect Workload logs

eloycoto:
This PR adds a way to send logs to a Syslog server, or any other platform that implements the interface in the future.

The main goal is to read Pod logs and store them on an intermediate struct (Fifolog) that buffers (With limits defined by a user) and sends it to the given server with retries, etc.. 

Users can define multiple Syslog servers at the device level, like: 

```
apiVersion: management.project-flotta.io/v1alpha1
kind: EdgeDevice
metadata:
  name: XXXX
  namespace: default
spec:
  log_collection:
    syslog:
      BufferSize: 10
      kind: syslog
      syslogConfig:
        address: dc1-log.acme.org:601
        proto: tcp
    importantLogs:
      BufferSize: 1000
      kind: syslog
      syslogConfig:
        address: custom-syslog.acme.org:601
        proto: tcp
``` 

And assign log transport on the deploy side using this format: 

```
---
apiVersion: management.project-flotta.io/v1alpha1
kind: EdgeDeployment
metadata:
  name: random-deployment
spec:
  deviceSelector:
    matchLabels:
      app: foo
  type: pod
  log_collection: importantLogs
  pod:
    spec:
      containers:
        - name: random-server
          image: eloycoto/test
          ports:
            - containerPort: 8080
              hostPort: 9099
```

Fix: https://issues.redhat.com/browse/ECOPROJECT-423
RFC-DOC: https://docs.google.com/document/d/1_hWAmLxZWkivIPe_XcoR6xxomuHFTqw4aB_A30MYOUc/edit
Depends-on: https://github.com/project-flotta/flotta-operator/pull/113

### DIFF
--- a/internal/logs/log_container.go
+++ b/internal/logs/log_container.go
@@ -1,0 +1,94 @@
+package logs
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+)
+
+const (
+	elapsedTime = 100 * time.Millisecond
+)
+
+type LogEntry struct {
+	Time     time.Time
+	data     []byte
+	workload string
+}
+
+func NewLogEntry(data []byte, workload string) *LogEntry {
+	return &LogEntry{
+		data:     data,
+		Time:     time.Now(),
+		workload: workload,
+	}
+}
+
+func (l *LogEntry) Size() int {
+	return len(l.data)
+}
+
+func (l *LogEntry) String() string {
+	if l.data == nil {
+		return ""
+	}
+	return string(l.data)
+}
+
+func (l *LogEntry) GetWorkload() string {
+	return l.workload
+}
+
+type FIFOLog struct {
+	maxlen      int
+	data        []*LogEntry
+	lock        sync.Mutex
+	currentSize int
+}
+
+// NewFIFOLog returns a new FifoLog struct
+func NewFIFOLog(maxSize int) *FIFOLog {
+	return &FIFOLog{
+		data:   []*LogEntry{},
+		maxlen: maxSize,
+	}
+}
+
+// CurrentSize returns the buffer size.
+func (f *FIFOLog) CurrentSize() int {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	return f.currentSize
+}
+
+// Write fills the current buffer with the next LogEntry
+func (f *FIFOLog) Write(entry *LogEntry) error {
+	if entry == nil {
+		return errors.New("No log entry added")
+	}
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	// We have here a "soft" limit, current entry size can be bigger than
+	// f.maxlen but we only pass once.
+	if f.currentSize > f.maxlen {
+		return fmt.Errorf("Buffer is already full")
+	}
+	f.data = append(f.data, entry)
+	f.currentSize = f.currentSize + entry.Size()
+	return nil
+}
+
+// ReadLine reads the line and delete it from the current buffer.
+func (f *FIFOLog) ReadLine() (*LogEntry, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	if f.currentSize <= 0 {
+		return nil, nil
+	}
+	res := f.data[0]
+	f.data = f.data[1:]
+	f.currentSize = f.currentSize - res.Size()
+	return res, nil
+}
--- a/internal/logs/syslog_transport.go
+++ b/internal/logs/syslog_transport.go
@@ -1,0 +1,85 @@
+package logs
+
+import (
+	"errors"
+	"fmt"
+	"log/syslog"
+	"net"
+	"os"
+	"sync"
+	"time"
+)
+
+const (
+	priority           = syslog.LOG_INFO
+	elapsedTimeOnError = 100 * time.Millisecond
+)
+
+// We implement this way to force timestamp with the original ones.
+type syslogTransport struct {
+	server   string
+	protocol string
+	hostname string
+	conn     net.Conn
+	lock     sync.Mutex // guards conn
+}
+
+func NewSyslogServer(server string, protocol string, deviceID string) (*syslogTransport, error) {
+	proto := "tcp"
+	if protocol != "" {
+		proto = protocol
+	}
+	syslogInstance := &syslogTransport{
+		server:   server,
+		protocol: proto,
+		hostname: deviceID,
+		lock:     sync.Mutex{},
+	}
+	err := syslogInstance.Init()
+	if err != nil {
+		return nil, err
+	}
+	return syslogInstance, nil
+}
+
+func (s *syslogTransport) Init() error {
+	var err error
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.conn, err = net.Dial(s.protocol, s.server)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func (s *syslogTransport) WriteLog(logline *LogEntry) error {
+	s.lock.Lock()
+	hostname := s.hostname
+	s.lock.Unlock()
+
+	logEntry := fmt.Sprintf("<%d>%s %s %s[%d]: %s\n",
+		priority, logline.Time.Format(time.RFC3339), hostname,
+		logline.workload, os.Getpid(), logline.String())
+	return s.writeAndRetry(logEntry, 10)
+}
+
+func (s *syslogTransport) writeAndRetry(data string, retryTimes int) error {
+	if retryTimes < 0 {
+		return errors.New("cannot send message")
+	}
+
+	s.lock.Lock()
+	if s.conn != nil {
+		_, err := fmt.Fprintf(s.conn, "%s", data)
+		s.lock.Unlock()
+		if err != nil {
+			time.Sleep(elapsedTimeOnError)
+			return s.writeAndRetry(data, retryTimes-1)
+		}
+		return nil
+	}
+	s.lock.Unlock()
+	return s.writeAndRetry(data, retryTimes-1)
+}
--- a/internal/logs/transport.go
+++ b/internal/logs/transport.go
@@ -1,5 +1,9 @@
 package logs
 
+const (
+	SyslogTransport = "syslog"
+)
+
 type transport interface {
-	WriteLog(logline *LogEntry, workload string) error
+	WriteLog(logline *LogEntry) error
 }
--- a/internal/logs/transport.go
+++ b/internal/logs/transport.go
@@ -1,0 +1,5 @@
+package logs
+
+type transport interface {
+	WriteLog(logline *LogEntry, workload string) error
+}
--- a/internal/logs/workload.go
+++ b/internal/logs/workload.go
@@ -1,0 +1,201 @@
+package logs
+
+import (
+	"context"
+	"time"
+
+	"git.sr.ht/~spc/go-log"
+	"github.com/project-flotta/flotta-device-worker/internal/workload"
+	"github.com/project-flotta/flotta-device-worker/internal/workload/podman"
+	"github.com/project-flotta/flotta-operator/models"
+)
+
+type transportLog struct {
+	transport transport
+	log       *FIFOLog
+	cancel    context.CancelFunc
+}
+
+func (t *transportLog) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.cancel = cancel
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				entry, err := t.log.ReadLine()
+				if err != nil {
+					log.Error("error on read log info", err)
+				}
+				if entry == nil {
+					time.Sleep(elapsedTime)
+				} else {
+					err = t.transport.WriteLog(entry)
+					if err != nil {
+						// cannot send, pushed again to the queue.
+						err := t.log.Write(entry)
+						if err != nil {
+							log.Error("cannot return log entry to the buffer log after transport failed", err)
+						}
+					}
+				}
+			}
+		}
+	}()
+}
+
+func (t *transportLog) Stop() {
+	if t.cancel != nil {
+		t.cancel()
+	}
+}
+
+type WorkloadsLogsTarget struct {
+	deviceID        string
+	transports      map[string]transportLog
+	wwriter         map[string]*WorkloadWriter // key is the workload name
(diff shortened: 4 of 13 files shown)


internal/logs/transport.go
```diff
@@ -0,0 +1,9 @@
+package logs
+
+const (
+	SyslogTransport = "syslog"
+)
+
+type transport interface {
+	WriTeLog(logline *LogEntry) error
```
jakub-dzon, line 8: T->t

internal/logs/log_container.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (l *LogEntry) String() string {
+	if l.data == nil {
+		return ""
+	}
+	return string(l.data)
+}
+
+func (l *LogEntry) GetWorkload() string {
```
jakub-dzon, line 39: What is the reason for having a getter instead of exported field?

internal/logs/syslog_transport.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (s *syslogTransport) Init() error {
+	hostname, _ := os.Hostname()
+	var err error
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.hostname = hostname
+	s.conn, err = net.Dial("tcp", s.server)
```
jakub-dzon, line 45: Isn't protocol configurable here: https://github.com/project-flotta/flotta-operator/blob/69cbc38035b7528216e064220e3047b3fa73d54c/swagger.yaml#L193 ?

internal/logs/syslog_transport.go
```diff
@@ -0,0 +1,80 @@
...
+	defer s.lock.Unlock()
+
+	if s.conn != nil {
+		_, err := fmt.Fprintf(s.conn, "%s", data)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	return s.writeAndRetry(data, retryTimes-1)
```
jakub-dzon, line 79: Why recursively?

jakub-dzon: It would be also good to pause between retries - exponential backoff maybe?

internal/logs/workload_writer.go
```diff
@@ -0,0 +1,60 @@
...
+
+func NewWorkloadWriter(workload string, manager *workload.WorkloadManager) *WorkloadWriter {
+	res := &WorkloadWriter{
+		workloadName: workload,
+		manager:      manager,
+	}
+	return res
+}
+
+func (ww *WorkloadWriter) SetTarget(name string, target *FifoLog) {
```
jakub-dzon, line 26: It does not seem to be called in other context than right after WorkloadWriter is created; can't that be a part of the constructor?

internal/logs/workload.go
```diff
@@ -0,0 +1,199 @@
...
+				}
+				if entry == nil {
+					time.Sleep(elapsedTime)
+				} else {
+					err = t.transport.WriTeLog(entry)
+					if err != nil {
+						// cannot send, pushed again to the queue.
+						err := t.log.Write(entry)
+						if err != nil {
+							log.Error("cannot return log entry to the buffer log after transport failed", err)
```
jakub-dzon, line 40: In that case that log entry is lost. In case of communication problems of the device, when transport can't connect for long time we might lose big chunks of logs. Is there a way for dealing with that I don't see?

internal/logs/workload.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (t *transportLog) Stop() {
+	if t.cancel != nil {
+		t.cancel()
+	}
+}
+
+type WorkloadsLogsTarget struct {
+	transports      map[string]transportLog
+	workloads       []*WorkloadWriter
```
jakub-dzon, line 57: There is a lot of map-like operations against that slice (getting by name, removing by name) - can it be a map and have O(1) instead of O(n) for these operations?

jakub-dzon (changes requested):

internal/logs/workload.go
eloycoto: I did as O(1) but I need to make complex structs, at the end, the number of workloads will be small(<10) so I think that it's much better dealing in this way than in the other way. 

This is my personal opinion, and by the kind of devices that we're supposed to manage

internal/logs/syslog_transport.go
eloycoto: ouch, good catch!

internal/logs/workload.go
eloycoto: we pushed again to the queue, and it has a buffer. Maybe in future we can add a fifolog/failed entry to a file, and sync when connectivity is off. This is the best effort and buffers size.

internal/logs/workload_writer.go
eloycoto: Want to make it simpler, we heavily use the constructor with a ton of params, and in this way, for me, it's a bit easier to understand. 

Also, this way I was also thinking on update transports in case of changes in the future.

internal/logs/log_container.go
```diff
@@ -0,0 +1,94 @@
...
+func NewFifoLog(maxSize int) *FifoLog {
+	return &FifoLog{
+		data:   []*LogEntry{},
+		maxlen: maxSize,
+	}
+}
+
+// CurrentSize returns the buffer size.
+func (f *FifoLog) CurrentSize() int {
+	f.lock.Lock()
```
machacekondra, line 60: What is the reason for this lock?

internal/logs/workload_writer.go
```diff
@@ -0,0 +1,60 @@
...
+	ww.target = target
+	ww.targetName = name
+}
+
+func (ww *WorkloadWriter) Write(data []byte) (n int, err error) {
+	entry := NewLogEntry(data, ww.workloadName)
+	if ww.target == nil {
+		return 0, fmt.Errorf("cannot write to target")
```
gciavarrini, line 34: In this case the problem is that there is not FIFOLog available, WDYT to make the log a little clearer?

internal/logs/workload_writer.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	err = ww.target.Write(entry)
+	if err != nil {
+		return 0, err
```
gciavarrini, line 38: Here I would add the "cannot write to target" to the `err`.

internal/logs/log_container.go
```diff
@@ -0,0 +1,94 @@
...
+		return ""
+	}
+	return string(l.data)
+}
+
+func (l *LogEntry) GetWorkload() string {
+	return l.workload
+}
+
+type FifoLog struct {
```
gciavarrini, line 43: I think that variable names that contains acronyms probably should be capitalized (FifoLog -> FIFOLog)

internal/logs/syslog_transport.go
```diff
@@ -0,0 +1,80 @@
...
+
+	logEntry := fmt.Sprintf("<%d>%s %s %s[%d]: %s\n",
+		priority, logline.Time.Format(time.RFC3339), hostname,
+		logline.workload, os.Getpid(), logline.String())
+	return s.writeAndRetry(logEntry, 10)
+}
+
+func (s *syslogTransport) writeAndRetry(data string, retryTimes int) error {
+	if retryTimes < 0 {
+		return errors.New("Cannot send message")
```
gciavarrini, line 66: Cannot -> cannot

internal/logs/log_container.go
eloycoto: Just in case that something changes in the future, like namespace-workload or something like this.

internal/logs/log_container.go
eloycoto: Write is called from another goroutine(WorkloadStarted) and Readline from target (Another goroutine) Current size is moved increased/decreased on each function, so lock is needed.

internal/logs/workload.go
jakub-dzon: IMO [these](https://github.com/project-flotta/flotta-device-worker/pull/101/files#diff-b5eb0d3de573bbf6fc6b46b3b42057737eb1b7c236b49b381231b7198e6aac9bR85-R94) 9 lines can be replaced with 3
and actually, why doesn't it use https://github.com/project-flotta/flotta-device-worker/pull/101/files#diff-b5eb0d3de573bbf6fc6b46b3b42057737eb1b7c236b49b381231b7198e6aac9bR97 ?

internal/logs/workload.go
jakub-dzon: But the buffer has limited size and can be full very quickly: https://github.com/project-flotta/flotta-device-worker/pull/101/files#diff-b642ccca5637f0ce9e7df1edc0d12111f67b1604b13bb89edc63830ba147ab9bR76.

jakub-dzon (commented):
How do we guarantee that we don't send the same logs after the agent is restarted - in different scenarios: just the device worker is restarted or the whole device is restarted.

eloycoto:
>How do we guarantee that we don't send the same logs after the agent is restarted - in different scenarios: 
> just the device worker is restarted 

In this scenario, logs in buffer will be lost. but when we retrieve logs is from that time, no more lines.  

> or the whole device is restarted.

The same as before, the logs are only retrieve since logs started.

internal/logs/workload.go
eloycoto: At the end, you need to loop over the deletetarget, but I'll change.

jakub-dzon:
> > How do we guarantee that we don't send the same logs after the agent is restarted - in different scenarios:
> > just the device worker is restarted
> 
> In this scenario, logs in buffer will be lost. but when we retrieve logs is from that time, no more lines.
> 
OK - and that's because https://github.com/project-flotta/flotta-device-worker/blob/43baf6f54980bd3d519bf9de638669fa25283112/internal/workload/podman/podman.go#L387 is set to 1?

internal/logs/workload.go
eloycoto: We can extend in future, users have the way to define how big the buffer can be, as defined in the doc. I'm happy to move to file, but what happens if the device is running without internet for a month and the log is duplicated and at the end the disk is full, it's a bit of chicken-egg problem, and at the end, it'll lose log files. 

The source of truth is always the device, what we proposed here is a best-effort solution IMHO.

eloycoto:
yes, it's because of that, you can duplicate one line for sure too, but I think that it's what makes sense.

internal/logs/workload.go
jakub-dzon: We can pause reading from the log when the buffer is full until something is eventually sent successfully. In case of reboot while disconnected it makes no difference if the source logs are unprocessed or lost from memory. To prevent that, we might somehow preserve information about the last read and sent line - to restart reading log file from that place after restart.

eloycoto: In that case, you're going to lose the logs that are going to be rotated on the journald process(Roy is working on that), so at the end is a similar situation that you will lose logs.

internal/logs/syslog_transport.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	err := syslogInstance.Init()
+	if err != nil {
+		return nil, err
+	}
+	return syslogInstance, nil
+}
+
+func (s *syslogTransport) Init() error {
+	hostname, _ := os.Hostname()
```
masayag, line 44: what if hostname isn't set for the device and it returns `localhost`?
also, should this error be ignored?

internal/logs/workload.go
```diff
@@ -0,0 +1,199 @@
...
+	return res
+}
+
+func (w *WorkloadsLogsTarget) GetCurrentWorkloads() []*WorkloadWriter {
+	return w.workloads
+}
+
+func (w *WorkloadsLogsTarget) deleteTargetTransport(name string) {
+	delete(w.transports, name)
+	temp_workloads := []*WorkloadWriter{}
```
masayag, line 85: nit: s/temp_workloads/tempWriters

internal/logs/workload.go
```diff
@@ -0,0 +1,199 @@
...
+	transports      map[string]transportLog
+	workloads       []*WorkloadWriter
+	workloadsConfig map[string]string
+	workloadManager *workload.WorkloadManager
+}
+
+func NewWorkloadsLogsTarget(manager *workload.WorkloadManager) *WorkloadsLogsTarget {
+	return &WorkloadsLogsTarget{
+		transports:      map[string]transportLog{},
+		workloads:       []*WorkloadWriter{},
```
masayag, line 65: the name a bit confusing, since reading the line below `w.workloads[i].Stop()` caused me to think we stopped a running workload.
I think it should be renamed to `writers` (or go with full name `workloadWriters`)

internal/logs/syslog_transport.go
```diff
@@ -0,0 +1,86 @@
...
+	var err error
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.hostname = hostname
+	s.conn, err = net.Dial(s.protocol, s.server)
```
masayag, line 51: how many open connection do we expect to have? Is it a single connection per LogCollection that will be used by multiple writers per workload?

internal/logs/syslog_transport.go
eloycoto: umm, thinking about this, maybe we can set the deviceID here.

internal/logs/syslog_transport.go
eloycoto: One connection per each LogCollection, where multiple writers can write too.

internal/logs/workload.go
```diff
@@ -0,0 +1,199 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				entry, err := t.log.ReadLine()
+				if err != nil {
+					log.Error("Error on read log info", err)
+				}
+				if entry == nil {
+					time.Sleep(elapsedTime)
```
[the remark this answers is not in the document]
eloycoto: is to read, so exp backoff is not the best thing here. It's elapsed time as normal.

internal/logs/syslog_transport.go
eloycoto: fixed.

eloycoto:
@jakub-dzon @machacekondra @masayag @gciavarrini, all comments have been addressed. Please have a look.

gciavarrini (approved):
lgtm

masayag (approved):

jakub-dzon (approved):

PR #102, opened 2022-03-01
Turn OS management off when rpm-ostree is not available

jakub-dzon:
This PR adds functionality of disabling OS-management features when there is no `rpm-ostree` tool installed.
It addresses issues reported in https://issues.redhat.com/browse/ECOPROJECT-588.


masayag (approved):

internal/os/os_exec_commands.go
```diff
@@ -21,13 +21,20 @@ type OsExecCommands interface {
...
+type osExecCommands struct{}
 
-func NewOsExecCommands() OsExecCommands{
+func NewOsExecCommands() OsExecCommands {
 	return &osExecCommands{}
 }
 
-func (o *osExecCommands) RpmOstreeStatus() ([]byte, error){
+func (o *osExecCommands) IsRpmOstreeAvailable() bool {
+	cmd := exec.Command("which", "rpm-ostree")
```
eloycoto, line 31: is which always present? I remember having an issue similar to this, but I do not recall where :\

eloycoto (commented):
Just a minor comment, but I love this one!

internal/os/os_exec_commands.go
jakub-dzon: Done - replaced with `rpm-ostree status`.

bardielle (approved):

PR #104, opened 2022-03-03
Tolerate missing OS configuration

jakub-dzon:
This PR resolves problem described in https://issues.redhat.com/browse/ECOPROJECT-587. And allows for OS configuration to be missing (which is the case before the device is registered).

### DIFF
--- a/internal/os/os.go
+++ b/internal/os/os.go
@@ -103,13 +103,14 @@ func (o *OS) Init(config models.DeviceConfigurationMessage) error {
 }
 
 func (o *OS) Update(configuration models.DeviceConfigurationMessage) error {
+	newOSInfo := configuration.Configuration.Os
+	if newOSInfo == nil {
+		log.Debug("No OS management configuration. Not updating.")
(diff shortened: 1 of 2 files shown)


bardielle (approved):

internal/os/os.go
```diff
@@ -103,14 +103,15 @@ func (o *OS) Init(config models.DeviceConfigurationMessage) error {
 }
 
 func (o *OS) Update(configuration models.DeviceConfigurationMessage) error {
+	newOSInfo := configuration.Configuration.Os
```
eloycoto, line 106: Should we implement this on Init() observer and skip the Update here?

jakub-dzon: I think it should be as it is - we want to protect `Update` from `nil` OS config in any case and don't report error because of that - not only for `Init`

machacekondra (approved):

pkliczewski (approved):

masayag:
LGTM

PR #107, opened 2022-03-08
Refactor deregistration

jakub-dzon:
This PR simplifies deregistration process and makes it easier to include new components in it.


bardielle (approved):

gciavarrini (approved):

eloycoto (commented):
I'm ok with changes, but shouldn't we use config/Observer method, and add a way method that it's Deregister and called from there? It will be simpler IMHO. 

So Observer will be 
Init()
Update()
Unregister()

Looks more natural for me.

bardielle:
@eloycoto I'm not sure I agree with you,
The observer is part of the configuration, but the deregistration logic is part of the Registration package.
Also, I don't think that all of the configuration entities need to implement a deregistration method.

masayag (approved):

masayag:
> So Observer will be
> Init()
> Update()
> Unregister()

The configmanager itself which is part of deregistration isn't an observer.
I'm going to merge this PR since it improves current state, but we can think of the pattern suggested by you.
Need to verify that there is no demand to coordinate the deregistration flow. @jakub-dzon WDYT?

jakub-dzon:
I agree with Danielle - configuration observers and components that need deregistration are different things and I wouldn't mix the two things. 
Also - if we used that interface, we should start (or finish) deregistration from (with) the configuration manager, which is not the case now (we start deregistration with workloads).

PR #108, opened 2022-03-08
Avoid duplicate configuration updates at the cost of increased contention

jakub-dzon:
This PR adds locking on handling of configuration messages to avoid unnecessary duplicate configuration updates.
Issue has been described in https://issues.redhat.com/browse/ECOPROJECT-284

### DIFF
--- a/internal/server/deviceserver.go
+++ b/internal/server/deviceserver.go
@@ -3,8 +3,6 @@ package server
 import (
 	"context"
 	"encoding/json"
-	"sync"
-
 	"git.sr.ht/~spc/go-log"
 	configuration2 "github.com/project-flotta/flotta-device-worker/internal/configuration"
 	registration2 "github.com/project-flotta/flotta-device-worker/internal/registration"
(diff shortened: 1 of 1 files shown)


internal/configuration/configuration.go
```diff
@@ -107,11 +107,12 @@ func (m *Manager) GetSecrets() models.SecretList {
 }
 
 func (m *Manager) Update(message models.DeviceConfigurationMessage) error {
-	m.lock.RLock()
+	m.lock.Lock()
+	defer m.lock.Unlock()
```
eloycoto, line 111: I highly disagree on this one TBH, because in this way if the config is finished, the config is locked on all observer, and it'll block GetDeviceID, GetSecrets, GetWorkloads functions if the observers use those ones.

eloycoto (changes requested):

internal/configuration/configuration.go
jakub-dzon: Hmmm, you're right; we may run into deadlocks in some cases... I forgot that go locks are not reentrant.

jakub-dzon:
@eloycoto WDYT now? It should not cause any deadlocks/race conditions and if configuration is the same, it should return pretty quickly.

internal/server/deviceserver.go
```diff
@@ -32,6 +34,9 @@ func (s *deviceServer) Send(ctx context.Context, d *pb.Data) (*pb.Receipt, error
 		if err != nil {
 			log.Warnf("cannot unmarshal message. DeviceID: %s; err: %v", s.configManager.GetDeviceID(), err)
 		}
+		// Don't allow for parallel configuration updates if one of them takes longer.
+		s.updateLock.Lock()
```
eloycoto, line 38: this lock will be "hard" because Update() can take a bit of time, and it's not a good way to manage it. Maybe what we need here is a waitinggroup, no?

https://gobyexample.com/waitgroups

jakub-dzon: The whole point of that lock is to allow for executing `Update` only once at a time - allowing simultaneous execution of `Update` is a problem. We can either use a lock and allow other updates to queue on it, queue the updates ourselves (here be dragons) or discard updates that come when an earlier update is in progress.

I don't see how wait group would help here. Potentially Try Lock behavior (for example https://github.com/subchen/go-trylock) would help in releasing goroutines early but I am _very_ hesitant about using 3rd party concurrency tools or writing them myself.:)

jakub-dzon: I see an option with atomic functions: use atomic-guarded `int32` as a try-lock:
```go
type deviceServer struct {
	locker              int32
}

  ...
  if swapped := atomic.CompareAndSwapInt32(&s.locker, 0, 1); swapped {
    err = s.configManager.Update(deviceConfigurationMessage)
    atomic.StoreInt32(&s.locker, 0)
  }
  // Ignore update
  ...
  return &pb.Receipt{}, nil
}

```

masayag: IMO, if we received a message to the device, we should process it.
With our current API, this is the behavior that we should aim for.
Unless we'll break the API into more specific actions.
The reason not to miss any update is the risk of missing an action, especially if we're going to to use device-configuration for running ansible playbooks.

jakub-dzon: With current ansible support design we can distinguish between ansible playbook messages and configuration updates and only ignore said configuration updates. 
I would prefer to lock and process all of the updates a bit later. That said I also don't think that skipping a configuration update would be a problem - we always receive full expected configuration and eventually we will process the most recent one. 
Of course there are corner cases - someone enables some setting for a moment and then changes it back to the original setting and the temporary configuration change might be missed by the device. Taking into account that we check for configuration updates in intervals, we can face the same situation now.

IMO we can safely go either way and I lean towards my original implementation - full lock.

jakub-dzon:
@masayag, @eloycoto - please have a look again; I think I found a way to fix the problem by avoiding explicit locking and update loss.

eloycoto (approved):

PR #110, opened 2022-03-14
Device: Implement MTLS registration 

eloycoto:
This PR adds a way to register a new device with the operator using the given Registration certificates and request a new certificate using the certificate signed request. 

On getting an HTTP http.StatusUnauthorized, on dispatcher.Send it'll trigger another Registration process, aka renew certificates.

It also updatse Yggdrasil version in a separate commit.  

 
Fix ECOPROJECT-399


cmd/device-worker/main.go
```diff
@@ -78,7 +78,7 @@ func main() {
 		log.Fatal(err)
 	}
 	if !r.GetRegistered() {
-		log.Fatal("handler registration failed")
+		log.Fatal("yggdrasil registration failed")
```
pkliczewski, line 81: Why do we need `yggdrasil` in this log entry? Maybe we could use `Registration failed`?

internal/registration/registration.go
```diff
@@ -82,30 +121,77 @@ func (r *Registration) registerDeviceWithRetries(interval int64) {
...
 		Content:   content,
 		Directive: "registration",
 	}
-
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
 
 	// Call "Send"
-	if _, err := r.dispatcherClient.Send(ctx, data); err != nil {
+	response, err := r.dispatcherClient.Send(ctx, data)
```
pkliczewski, line 154: How are we going to handle registration approval? There could a lot of time before admin approves registration.

cmd/device-worker/main.go
```diff
@@ -148,7 +148,14 @@ func main() {
 	hbs := heartbeat2.NewHeartbeatService(dispatcherClient, configManager, wl, &hw, dataMonitor, deviceOs)
 	configManager.RegisterObserver(hbs)
 
-	reg := registration2.NewRegistration(deviceId, &hw, dispatcherClient, configManager, wl)
+	reg, err := registration2.NewRegistration(deviceId, &hw, dispatcherClient, configManager, wl)
+	if err != nil {
+		log.Fatalf("Cannot started registration process:  DeviceID: %s; err: %v", deviceId, err)
```
pkliczewski, line 153: s/Cannot started/Cannot start

internal/heartbeat/heartbeat.go
```diff
@@ -126,7 +163,7 @@ func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) error {
 }
 
 func (s *Heartbeat) getInterval(config models.DeviceConfiguration) int64 {
-	var interval int64 = 60
+	var interval int64 = 10
```
pkliczewski, line 166: What do we need to change?

internal/registration/registration.go
```diff
@@ -131,3 +217,12 @@ func (r *Registration) Deregister() error {
 	r.registered = false
 	return errors
 }
+
+type YGGDresponse struct {
```
pkliczewski, line 221: why not to have it next to other types at the top of the file?

internal/heartbeat/heartbeat.go
eloycoto: Ouch, good catch, leftover :sweat:

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (c *ClientCert) IsRegisterCert() (bool, error) {
+	if c.certGroup == nil {
+		return false, errors.New("Certificates are not imported")
+	}
+
+	cert := c.certGroup.GetCert()
+	if cert == nil {
+		return true, errors.New("No valid cert")
```
bardielle, line 63: why do you return true here? I think it should be false
it means that the device has't  certificates

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return cert.Subject.CommonName == mtls.CertRegisterCN, nil
+}
+
+func (c *ClientCert) Renew(deviceID string, key crypto.PrivateKey) ([]byte, []byte, error) {
```
bardielle, line 69: It's not really clear what should be the 2 firsts return parameters - 
Maybe we can use a struct here?

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+	keyPEM, err := c.certGroup.MarshalKeyToPem(key)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot encode new key: %v", err)
+	}
+
+	return certPEM, keyPEM.Bytes(), nil
+}
+
+func (c *ClientCert) CreateDeviceCerts(deviceID string) ([]byte, []byte, error) {
+	key, err := c.certGroup.GetNewKey()
```
bardielle, line 96: The same here - I think we can use a struct for the return parameters

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return fmt.Errorf("failing on importing certificate: %v", err)
+	}
+
+	err = ioutil.WriteFile(c.certPath, cert, 0600)
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(c.keyPath, key, 0600)
```
bardielle, line 120: I think it should be an atomic functionality - 
If something went wrong here -with writing the keyPath you will have a new certPath but the keyPath will be the old one 

Maybe keep the old pair in a backup file

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return cert.Subject.CommonName == mtls.CertRegisterCN, nil
+}
+
+func (c *ClientCert) Renew(deviceID string, key crypto.PrivateKey) ([]byte, []byte, error) {
+	var csrTemplate = x509.CertificateRequest{
+		Subject: pkix.Name{
+			CommonName: deviceID,
+			// Operator will add metadata on this subject, like namespace
```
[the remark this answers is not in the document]
eloycoto: Source of truth is the operator, it'll be done in the operator.

internal/registration/certs.go
eloycoto: If you look at the next lines, if key cannot be written it restore the old cert.

bardielle: I missed that it's using the certGroup as a backup

internal/registration/certs.go
masayag: don't think about struct, named parameters for the returned value is a simpler option (which I don't think is needed here due to line 92). However, documenting the function can be helpful for that.

internal/registration/registration.go
masayag: nit: s/YGGDresponse/YGGDResponse

internal/registration/registration.go
```diff
@@ -131,3 +217,12 @@ func (r *Registration) Deregister() error {
...
+	// StatusCode response
+	StatusCode int
+	// Response Body
+	Body json.RawMessage
+	// Mestadata added by the transport, in case of http are the headers
```
masayag, line 226: s/Mestadata/Metadata

internal/registration/registration.go
```diff
@@ -82,30 +121,77 @@ func (r *Registration) registerDeviceWithRetries(interval int64) {
...
+
+	parsedContent, ok := message.Content.(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("Cannot parse message content, content=%+v", message.Content)
+	}
+
+	cert, ok := parsedContent["certificate"]
+	if !ok {
+		return fmt.Errorf("Cannot retrieve certificate from parsedResponse, content=%+v", message.Content)
+	}
```
masayag, line 182: Do we wish to print the entire error content to log?couldn't it contain sensitive data?

internal/registration/registration.go
```diff
@@ -82,30 +121,77 @@ func (r *Registration) registerDeviceWithRetries(interval int64) {
...
+
+	parsedCert, ok := cert.(string)
+	if !ok {
+		return fmt.Errorf("Cannot parse certificate from response.Content, content=%+v", message.Content)
+	}
+
+	err = r.clientCert.WriteCertificate([]byte(parsedCert), key)
+	if err != nil {
+		fmt.Println("Failed to write certificate!,", err)
```
masayag, line 191: pls replace with `log.Errorf`

masayag (changes requested):

internal/heartbeat/heartbeat.go
```diff
@@ -97,6 +101,39 @@ func NewHeartbeatService(dispatcherClient pb.DispatcherClient, configManager *cf
...
+	}
+
+	var parsedResponse registration.YGGDresponse
+	err = json.Unmarshal(response.Response, &parsedResponse)
+	if err != nil {
+		return err
+	}
+
+	// If it's already expired the cert need to be renewed
+	if parsedResponse.StatusCode != http.StatusUnauthorized {
```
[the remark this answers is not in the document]
eloycoto: So, if something different than 401, it'll return, if 401, retrigger the registration.

internal/registration/registration.go
eloycoto: The device will be "registerted" but now allowed to pull info/config, etc.. ;-)

cmd/device-worker/main.go
eloycoto: this is registration from device-worker to yggdrasil server.

internal/registration/registration.go
eloycoto: Discussed offline, let's move this in another task https://issues.redhat.com/browse/ECOPROJECT-219

internal/registration/certs.go
```diff
@@ -0,0 +1,131 @@
...
+
+	keyPEM, err := c.certGroup.MarshalKeyToPem(key)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot encode new key: %v", err)
+	}
+
+	return certPEM, keyPEM.Bytes(), nil
+}
+
+func (c *ClientCert) CreateDeviceCerts(deviceID string) ([]byte, []byte, error) {
```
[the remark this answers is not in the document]
eloycoto: As previous comment, documment it ;-)

cmd/device-worker/main.go
```diff
@@ -148,7 +148,14 @@ func main() {
 	hbs := heartbeat2.NewHeartbeatService(dispatcherClient, configManager, wl, &hw, dataMonitor, deviceOs)
 	configManager.RegisterObserver(hbs)
 
-	reg := registration2.NewRegistration(deviceId, &hw, dispatcherClient, configManager, wl)
+	reg, err := registration2.NewRegistration(deviceId, &hw, dispatcherClient, configManager, wl)
```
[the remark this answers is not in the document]
eloycoto: Done

go.sum
```diff
@@ -1665,8 +1666,8 @@ github.com/quasilyte/go-consistent v0.0.0-20190521200055-c6f3937de18c/go.mod h1:
 github.com/quasilyte/go-ruleguard v0.2.0/go.mod h1:2RT/tf0Ce0UDj5y243iWKosQogJd8+1G3Rs2fxmlYnw=
 github.com/quasilyte/regex/syntax v0.0.0-20200407221936-30656e2c4a95/go.mod h1:rlzQ04UMyJXu/aOvhd8qT+hvDrFpiwqp8MRXDY9szc0=
 github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a/go.mod h1:bCqnVzQkZxMG4s8nGwiZ5l3QUCyqpo9Y+/ZMZ9VjZe4=
-github.com/redhatinsights/yggdrasil v0.0.0-20220210193750-d696ee7a54bb h1:J+WMR8ee4Yrkg7nCqXOaE8oPyKbhdc7AbVSokEwnKMA=
-github.com/redhatinsights/yggdrasil v0.0.0-20220210193750-d696ee7a54bb/go.mod h1:pzYHLc9dg7BTax4FlcgVjdrN7SWHrmXCaQ2kUXWUGK8=
+github.com/redhatinsights/yggdrasil v0.0.0-20220309180936-d91759a7f332 h1:VPcx9axsKg7kifVVNcpF4qhfdc8BSQCrrFdfTdZJ0OE=
```
pkliczewski, line 1669: Do we know when yggdrasil will be release with the changes?

eloycoto: No timeline yet, but with these changes and also with all HTTP transport ones. I'll investigate :-)

internal/heartbeat/heartbeat.go
```diff
@@ -97,6 +104,34 @@ func NewHeartbeatService(dispatcherClient pb.DispatcherClient, configManager *cf
 	}
 }
 
+func (s *Heartbeat) send(data *pb.Data) error {
+	// Just make sure that all send finished
+	s.wg.Wait()
```
jakub-dzon, line 109: Why not a mutex?

eloycoto: Mutex on external dependencies can be super evil, at the end here, we only need to wait for anything to finish.

jakub-dzon: Actually, when I look at it, there is a deadlock, I think: you call `wg.Wait()` before there was any `wg.Add` so no go routine is unable to proceed to executing the method.

jakub-dzon: OK, scratch that:D If the there was no `Add`, there is nothing to  wait for and it should go through... :facepalm:

eloycoto: https://go.dev/play/p/kcwSfkjTfSS

jakub-dzon: In a corner case there might be more go routines sending than 1:
- three go routines: A, B, C
- A increments the counter in wg and starts sending
- B comes and stops at `Wait`
- C comes and stops at `Wait`
- A finishes and calls `wg.Done`
- B and C are released, increment the counter twice and both start sending.

Is that expected? If no, then why not mutex, which will block concurrent execution?

bardielle (approved):

jakub-dzon (approved):

PR #115, opened 2022-03-20
metrics remote write

ydayagi:
send metrics from device to remote server

https://issues.redhat.com/browse/ECOPROJECT-557


internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	Url     string // empty URL means that feature is disabled
+	Timeout time.Duration
+}
+
+func NewRemoteWrite(dataDir, deviceID string, tsdbInstance API) *RemoteWrite {
+	waitInterval, _ := time.ParseDuration(os.Getenv("REMOTEWRITE_WAIT_INTERVAL"))
+	if waitInterval == 0 {
+		waitInterval = defaultWaitInterval
+	}
+	RangeDuration, _ := time.ParseDuration(os.Getenv("REMOTEWRITE_RANGE_DURATION"))
```
eloycoto, line 56: Overall, I think that all of those env variables should be part of the main device config, no? so users can update from config.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	lastWriteBytes, err := ioutil.ReadFile(newRemoteWrite.lastWriteFile)
+	if err == nil {
+		if len(lastWriteBytes) != 0 {
+			i, err := strconv.ParseInt(string(lastWriteBytes), 10, 64)
+			if err == nil {
+				newRemoteWrite.LastWrite = time.Unix(i/1000000000, i%1000000000)
+			} else {
+				log.Error(err)
```
eloycoto, line 86: The log line should have more data, like why it happens.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	case remote.RecoverableError:
+		return RemoteRecoverableError{err}
+	default:
+		return err
+	}
+}
+
+func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
+	// the following two variables are only here until we implement reading form device configuration
+	serverURL := os.Getenv("REMOTEWRITE_SERVER_URL")
```
eloycoto, line 133: this shouldn't part of the config for the device? getting from the API?

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	go r.writeRoutine()
+	return err
+}
+
+// writeRoutine
+// Used as the goroutine function for handling ongoing writes to remote server
+func (r *RemoteWrite) writeRoutine() {
+	log.Infof("metric remote writer %s started", r.deviceID)
+	for {
+		r.Write()
```
eloycoto, line 183: what happens if this get stuck and no longer pushed data? Should be a timeout here with a context?

eloycoto (changes requested):

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+		time.Sleep(r.waitInterval)
+	}
+}
+
+// write
+// Writes all the metrics. Stops when either no more metrics or failure (after retries)
+func (r *RemoteWrite) Write() {
```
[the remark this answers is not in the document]
jakub-dzon: Yes, TSDB should remove old blocks according to configured retention policy.

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,207 @@
...
+		tsdbInstance.EXPECT().GetMetricsForTimeRange(midTime.Add(time.Millisecond), maxTime, true).Return(series, nil).Times(1)
+		writeClient.EXPECT().Write(gomock.Any(), gomock.Any()).Return(nil).Times(2)
+		remoteWrite.LastWrite = lastWrite
+		remoteWrite.Write()
+		Expect(remoteWrite.LastWrite).To(Equal(maxTime))
+	})
+
+	It("device label", func() {
+		minTime := time.Now()
+		maxTime := minTime
```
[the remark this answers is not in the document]
jakub-dzon: :+1: from me. For me it improves readability.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+			client: client,
+		}
+	}
+
+	return nil
+}
+
+func (r *RemoteWrite) Init(config models.DeviceConfigurationMessage) error {
+	err := r.Update(config)
+	go r.writeRoutine()
```
jakub-dzon, line 174: if the feature is disabled (https://github.com/project-flotta/flotta-device-worker/pull/115/files/3a8d1de8e000a338a61f1e57bc29c1924ab0e1ab#diff-51602231cf99bc104842e148b00d32af4ed8812fbf41d942e7b9ed9417ee048dR145), that go routine should not be started.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+		client, err := remote.NewWriteClient(r.deviceID, &remote.ClientConfig{
+			Timeout: model.Duration(r.config.Timeout),
+			URL: &config_util.URL{
+				URL: serverURL,
+			},
+		})
+		if err != nil {
+			return err
+		}
+		r.Client = &writeClient{
```
jakub-dzon, line 164: `r.Client` is accessed from another go routine (https://github.com/project-flotta/flotta-device-worker/pull/115/files/3a8d1de8e000a338a61f1e57bc29c1924ab0e1ab#diff-51602231cf99bc104842e148b00d32af4ed8812fbf41d942e7b9ed9417ee048dR180) and without constructs introducing happens-before relationship (i.e. locks or atomics), that change might not be visible in a non-deterministic way.

jakub-dzon (changes requested):

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+		rangeStart := r.LastWrite.Add(time.Millisecond) // TSDB time is in milliseconds
+		if rangeStart.Sub(minTSDB) < 0 {
+			rangeStart = minTSDB
+		}
+		rangeEnd := rangeStart.Add(r.RangeDuration)
+		if rangeEnd.Sub(maxTSDB) > 0 {
+			rangeEnd = maxTSDB
+		}
+
+		log.Infof("going to write metrics range %s(%d)-%s(%d). TSDB min max: %s(%d)-%s(%d)",
```
gciavarrini, line 215: Maybe a debug level is more appropriate here?

internal/metrics/remote_write.go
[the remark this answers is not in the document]
eloycoto: Then, on device restart it'll send duplicate the info to the remote-write, no?

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+		reqBytes, err := json.Marshal(writeRequest)
+		if err != nil {
+			log.Error("can not marshal prompb.WriteRequest to json", err)
+		}
+		log.Tracef("sending write request (lowest %s highest %s): %s", lowest.String(), highest.String(), string(reqBytes))
+	}
+
+	reqBytes, err := writeRequest.Marshal()
+	if err != nil {
+		log.Error("can not marshal prompb.WriteRequest", err)
```
gciavarrini, line 326: for consistency  `s/can not/cannot`

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+			// cause loop to break. skip this request cause error not recoverable
+			// e.g. already wrote these samples
+		}
+	}
+
+	return true
+}
+
+// toTimeSeries
+// converst to TimeSeries and returns lowest and highest timestamps
```
gciavarrini, line 360: /converst/convert/

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,207 @@
...
+	AfterEach(func() {
+		defer GinkgoRecover()
+		mockCtrl.Finish()
+		os.Remove(metrics.LastWriteFileName)
+	})
+
+	It("empty URL feature disabled", func() {
+		remoteWrite.Client = nil
+		remoteWrite.Write()
+	})
```
gciavarrini, line 44: There is no a `then` section here (I mean no `Expect()` statements). Is this test correct?

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+	It("new device", func() {
+		tsdbInstance.EXPECT().MaxTime().Return(time.Time{}).Times(1)
+		remoteWrite.Write()
```
gciavarrini, line 48: No `Expect` invocation after `remoteWrite.Write()`. Is it ok?

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	lastWriteFile        string
+	waitInterval         time.Duration
+	RangeDuration        time.Duration
+	RequestNumSamples    int
+	RequestRetryInterval time.Duration
+	Client               WriteClient
+}
+
+type RemoteWriteConfiguration struct {
+	Url     string // empty URL means that feature is disabled
```
pkliczewski, line 47: Let's make sure it is documented.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	// create configuration
+	r.config = RemoteWriteConfiguration{
+		Url:     serverURL,
+		Timeout: serverTimeout,
+	}
+
+	if r.config.Url == "" { // feature is disabled
+		return nil
```
pkliczewski, line 146: what do you think about logging it here?

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	// create client
+	if r.Client == nil {
+		serverURL, err := url.Parse(r.config.Url)
+		if err != nil {
+			return err
```
pkliczewski, line 153: adding a log here would be helpful during troubleshooting

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+func (r *RemoteWrite) Init(config models.DeviceConfigurationMessage) error {
+	err := r.Update(config)
+	go r.writeRoutine()
+	return err
+}
+
+// writeRoutine
+// Used as the goroutine function for handling ongoing writes to remote server
+func (r *RemoteWrite) writeRoutine() {
+	log.Infof("metric remote writer %s started", r.deviceID)
```
pkliczewski, line 181: should we add some debug info about time and interval?

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+
+		// add device label
+		for _, s := range series {
+			s.Labels[DeviceLabel] = r.deviceID
+		}
+
+		// write range
+		if len(series) != 0 {
+			if r.writeRange(series) {
```
pkliczewski, line 236: Should we use a data limit here? We do not want to saturate network after the device was disconnected for extended period of time.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	if err != nil {
+		log.Error("can not marshal prompb.WriteRequest", err)
+		return false
+	}
+
+	reqBytes = snappy.Encode(nil, reqBytes)
+
+	const numTries = 3
+	for try := 1; try <= numTries; try++ {
+		err = r.Client.Write(context.TODO(), reqBytes)
```
pkliczewski, line 334: Can we block here? What about timeout?

internal/metrics/remote_write.go
ydayagi: it is internal. not exposed to users and hopefully never will be changed. the env vars are more for development

internal/metrics/remote_write.go
ydayagi: we log the error. it has all the details. I do not know what the error is so I cannot add any relevant info.

internal/metrics/remote_write.go
ydayagi: in next story it will come from device config. see the comment in code

internal/metrics/remote_write.go
ydayagi: it can not get stuck. the code is non-blocking until we reach the HTTP client. there a context with timeout is created. the timeout is configured with the server URL from device config (at the moment env var)

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+			}
+		} else {
+			log.Info("metrics range is empty")
+		}
+
+		// store last write
+		r.LastWrite = rangeEnd
+		err = ioutil.WriteFile(r.lastWriteFile, strconv.AppendInt(nil, r.LastWrite.UnixNano(), 10), 0600)
+		if err != nil {
+			log.Error(err)
```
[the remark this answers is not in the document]
ydayagi: when the Write iteration starts we log the relevant range along with TSDB min and amx. IMO no need to repeat it here as well. this code is very unlikely to fail.

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: we let retention delete old metrics. we progress with the last write timestamp. so we never resend metrics again.

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,207 @@
...
+		tsdbInstance.EXPECT().GetMetricsForTimeRange(minTime, maxTime, true).Return(series, nil).Times(1)
+		writeClient.EXPECT().Write(gomock.Any(), gomock.Any()).Return(metrics.RemoteRecoverableError{}).Times(3)
+		remoteWrite.RequestRetryInterval = 0
+		lastWriteBefore := remoteWrite.LastWrite
+		remoteWrite.Write()
+		Expect(remoteWrite.LastWrite).To(Equal(lastWriteBefore))
+	})
+
+	It("request send failed once - recoverable error", func() {
+		minTime := time.Now()
```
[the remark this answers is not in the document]
ydayagi: not every line of code the repeats should be pushed into function

internal/metrics/remote_write_test.go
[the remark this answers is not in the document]
ydayagi: I don't like these comments and do not see why do we use them. IMO it does not describe the test and besides.....u already use Ginkgo to explain your test.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+				log.Infof("wrote metrics range %s(%d)-%s(%d)", rangeStart.String(), rangeStart.UnixNano(), rangeEnd.String(), rangeEnd.UnixNano())
+			} else {
+				return // all tries failed. we will retry later
+			}
+		} else {
+			log.Info("metrics range is empty")
+		}
+
+		// store last write
+		r.LastWrite = rangeEnd
```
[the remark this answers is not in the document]
ydayagi: only one goroutine is executed

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	Client               WriteClient
+}
+
+type RemoteWriteConfiguration struct {
+	Url     string // empty URL means that feature is disabled
+	Timeout time.Duration
+}
+
+func NewRemoteWrite(dataDir, deviceID string, tsdbInstance API) *RemoteWrite {
+	waitInterval, _ := time.ParseDuration(os.Getenv("REMOTEWRITE_WAIT_INTERVAL"))
```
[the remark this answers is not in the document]
ydayagi: I disagree. these are internal parameters. Mostly for development. env vars are the most used for such vars that are internal. You can easily add them to the device's TOML file. Even in real installations.

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,394 @@
...
+	switch err.(type) {
+	case remote.RecoverableError:
+		return RemoteRecoverableError{err}
+	default:
+		return err
+	}
+}
+
+func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
+	// the following two variables are only here until we implement reading form device configuration
```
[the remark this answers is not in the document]
ydayagi: exactly. we have a story for adding this configuration to device. but I did not want to mix the two stories.

internal/metrics/remote_write.go
ydayagi: most chances are that it will be enabled. if disabled then probably due to some temporary problem. so I think it is easier to start it and let it operate/not-operate based on config. In general, I'm quite certain that we have other routines that are started even when the feature is disabled. We also create quite a lot of goroutines. I don't think another one is should be a problem. anyhow, I'll handle it in the DeviceConfig story.

internal/metrics/remote_write.go
ydayagi: at the moment only one routine really accesses it. During Init Client is created and never changes. Update only tests whether it is nil for the sake of Init calling Update. When the story for DeviceConfig will be implemented we'll have to change it to atomic.

internal/metrics/remote_write.go
ydayagi: On one hand yes. On the other if we see that metrics are missing, how can we tell that the device sent them?  If no one else objects and you still think it should be debug then I'll be glad to change (less noise in log)

internal/metrics/remote_write_test.go
ydayagi: yes :-). I expect to have 0 calls to TSDB and the client.

internal/metrics/remote_write_test.go
ydayagi: the expect is before. Expect a call to MaxTime.

internal/metrics/remote_write.go
ydayagi: sure. I'll add a story for documentation.

internal/metrics/remote_write.go
ydayagi: whatever u say. IMO we will be flooding the log.

internal/metrics/remote_write.go
ydayagi: the caller will log the error.

internal/metrics/remote_write.go
ydayagi: I can add all the configuration. I just thought it floods the log. your call.

internal/metrics/remote_write.go
ydayagi: writeRange breaks the range into RequestMaxSamples (just like ACM does).

internal/metrics/remote_write.go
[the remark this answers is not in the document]
eloycoto: if device restart, the TSDB will be in there for a week, and the lasttime will be nil, no?

internal/metrics/remote_write.go
ydayagi: Client is created with timeout. The timeout is part of RemoteWriteConfiguration. It will eventually come from control plane as part of remote server config.

internal/metrics/remote_write.go
eloycoto: I highly disagree here, we need to think that it's device level, and maybe those devices are using solar panels, batteries, etc.. having another goroutine can consume more electricity and we should avoid it.

internal/metrics/remote_write.go
masayag: Since this is part of Update, we should probably not log it here.
However, what will be the behavior when this feature is enabled and then disabled?

internal/metrics/remote_write.go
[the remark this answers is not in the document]
eloycoto: Ohh, nothing, lastwrite is stored in a file, missed it.

internal/metrics/remote_write.go
eloycoto: well, it's a context.TODO() so no timeout at all: 

https://github.com/project-flotta/flotta-device-worker/pull/115/files/3a8d1de8e000a338a61f1e57bc29c1924ab0e1ab#diff-51602231cf99bc104842e148b00d32af4ed8812fbf41d942e7b9ed9417ee048dR334

internal/metrics/remote_write_test.go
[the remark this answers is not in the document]
masayag: I'd prefer keeping all tests consistent. The alternative of  // given // when // then is using By("...") to describe specific parts  of the test or other notation that will be reflected in the code and in the test output.

internal/metrics/remote_write.go
ydayagi: if u look at the Client code you'll see the timeout. Not our code.
untitled (5)

internal/metrics/remote_write_test.go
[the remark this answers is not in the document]
ydayagi: @eloycoto you're right.. will fix.

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: later when update is really used we will add atomic/lock. https://issues.redhat.com/browse/ECOPROJECT-555

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: whether we like it or not env vars are used extensively in device. it is linux after all. Even YGGDRASIL has the TOML with the env section. I didn;t have to make it configurable. I prefer removing it than changing it to anything else.

internal/metrics/remote_write.go
ydayagi: The devices won't be anything less than a reasonable computer. look at our code. it is more than a small device can handle.
if this goroutine executes once every 5 minute it's nothing. Besides, if that is the case then entire code should be reviewed. For example. The metric targets. each is a goroutine instead of one. These create context switching and that means that CPU registers are written with new data and that means consuming electricity. The same goes for the node exporter that uses its own process. The same goes for our over frequent GET configuration and heartbeats. Let's be reasonable. We're not talking about a device operated with batteries nor do we have any idea what kind of devices we have.
And I already wrote it would be handled in the Device Config story (https://issues.redhat.com/browse/ECOPROJECT-555). Cause there we'll finalize who creates what first.

internal/metrics/remote_write.go
ydayagi: https://issues.redhat.com/browse/ECOPROJECT-555

internal/metrics/remote_write.go
ydayagi: @pkliczewski @jakub-dzon @eloycoto @masayag what do u think? Gloria raised an important concern

eloycoto: I'm ok to get it on info.

masayag: @ydayagi Wouldn't logging only the failure be enough to debug this issue?
I mean if it reaches to line 215,  the device is about to write the data.
If it failed, there should be an error to record that.

internal/metrics/remote_write.go
ydayagi: we will disable the feature. I didn't do it in this PR. I will do it as part of https://issues.redhat.com/browse/ECOPROJECT-555. There we will add config from cluster/operator. At the moment Update does not really do anything.

internal/metrics/remote_write.go
ydayagi: Here's the client code
untitled (5)

internal/metrics/remote_write.go
ydayagi: https://issues.redhat.com/browse/ECOPROJECT-677

internal/metrics/remote_write.go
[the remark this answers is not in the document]
eloycoto: the thing is, these devices are in somewhere that maybe does not have SSH access or even physical access at all. If it's to be fixed, internal should be a const, if not part of the device model.

internal/metrics/remote_write.go
eloycoto: then a const, for dev you can update.

internal/metrics/remote_write.go
eloycoto: for me is a bit clear: `log.Error(err, "Device cannot parse last checkpoint value for remote writer")` than cannot convert atoi into int in the log that can comes from anywhere.

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: I prefer to keep the env vars. it is a great tool, it gives more than using consts and it does not bother anyone.

internal/metrics/remote_write.go
ydayagi: @eloycoto @jakub-dzon 
I think I confused you.
There's no enable/disable for this feature. It is always enabled. The epic handles setting the URL.
I added this option for development. Similar to OBC_AUTO_CREATE in 'make run' of operator. Only here I have the privilege to do it without changing the makefile.

internal/metrics/remote_write_test.go
ydayagi: I'll add

internal/metrics/remote_write.go
ydayagi: @masayag see my comment to eloy and jakub. There's no disable for this feature. only for development. Part of the epic is installing Thanos and setting the URL.

ydayagi: @masayag changed the comment

internal/metrics/remote_write_test.go
ydayagi: @gciavarrini done

internal/metrics/remote_write.go
ydayagi: @gciavarrini done

internal/metrics/remote_write.go
ydayagi: @gciavarrini done

internal/metrics/remote_write_test.go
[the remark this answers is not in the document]
ydayagi: done

internal/metrics/remote_write_test.go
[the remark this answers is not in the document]
ydayagi: done

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: @eloycoto done

internal/metrics/remote_write.go
ydayagi: @eloycoto done

internal/metrics/remote_write.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	lastWriteBytes, err := ioutil.ReadFile(newRemoteWrite.lastWriteFile)
+	if err == nil {
+		if len(lastWriteBytes) != 0 {
+			i, err := strconv.ParseInt(string(lastWriteBytes), 10, 64)
+			if err == nil {
+				newRemoteWrite.LastWrite = time.Unix(i/1000000000, i%1000000000)
+			} else {
+				log.Errorf("can not parse metrics last write from file %s. error: %s", newRemoteWrite.lastWriteFile, err.Error())
```
gciavarrini, line 86: s/can not/cannot/

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,253 @@
...
+		mockCtrl.Finish()
+		os.Remove(metrics.LastWriteFileName)
+	})
+
+	It("empty URL feature disabled", func() {
+		// given
+		remoteWrite.Client = nil
+
+		// when
+		remoteWrite.Write()
```
gciavarrini, line 48: I'm still not sure about this test. 
This test will fail only in case of `panic`, it doesn't guarantee that there are no invocation to TSBD and to the client.
I see the value of having a test that covers this scenario, I just don't think this is the right way.

@eloycoto WDYT?

internal/metrics/remote_write.go
eloycoto: The use cases are: 
- If a customer for X reasons has that ENV variable for any reason, the program will miss behaving, and also it's going to be hard to debug. . 
- Those devices can be in any place without any physical security, you don't want to have any env variable out there that it's not documented that can be a vector of attack.

So, if it's not going to be in the deviceConfig from the API, I wouldn't take an extra risk on that.

internal/metrics/remote_write_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+		// then
+		Expect(remoteWrite.LastWrite).To(Equal(maxTime))
+	})
+
+	It("continue from last write", func() {
+		// given
+		minTime := time.Now()
```
eloycoto, line 83: should be deleted.

internal/metrics/remote_write.go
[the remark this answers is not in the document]
eloycoto: One thing is that cool, and another thing is that should be used, I already explained here:

https://github.com/project-flotta/flotta-device-worker/pull/115/files#r831993255

internal/metrics/remote_write_test.go
eloycoto: I would add a comment that no gomock expected call and should be ok

ydayagi: 1. testing that no panic happens is very important.
2. it does test that no invocation of TSDB and client happens. It does that by not expecting any calls.

internal/metrics/remote_write.go
pkliczewski: We already log some info here. I think adding time/interval would help us diagnose issues

internal/metrics/remote_write.go
pkliczewski: ok, can we make it configurable?

internal/metrics/remote_write.go
pkliczewski: How do we make sure that timeout * `RequestMaxSamples` is shorter than `waitInterval`?

internal/metrics/remote_write_test.go
ydayagi: done

internal/metrics/remote_write.go
ydayagi: I am not sure what do u mean by timeout*RequestNumSamples. I will assume that u mean that 'time for all requests to server' is shorter than waitInterval.
The two are not related.
The waitInterval is a period of time to wait for sufficient data to be created in TSDB. This is AFTER we finish sending all the current data. So, one write iteration starts, it can take as much time as needed and then we wait 'waitInterval' for more data (notice that I do not use a ticker or timer but a 'sleep' between iterations).
I also use it as a period to wait for temporary errors to be resolved. This is because the interval is like infinity for the retries. Retry happens every X seconds and the interval in minutes.
I hope I was clear enough.

internal/metrics/remote_write.go
ydayagi: sure. at the moment it is treated as an internal config and set by env var. I can add it to the DeviceConfig. Go for it?

internal/metrics/remote_write.go
ydayagi: we have a lot of env vars just like these and we never raised these concerns. I think you took it too far.

gciavarrini (approved):

internal/metrics/remote_write.go
eloycoto: we have a lot of env vars on the operator, not in the device, which is a big difference for me.

internal/metrics/remote_write.go
jakub-dzon: The feature is disabled if the URL is empty and in that case the go routine should not be started. 
Besides, I think it is a very valid use case for users to opt-out from collecting metrics on the cluster side. They might not want to maintain unused Thanos instance for the sake of running edge devices.

jakub-dzon (changes requested):

internal/metrics/remote_write.go
ydayagi: if it is a valid case then we have to handle disabling the metrics in the device. at the moment we do not. TSDB starts at least one routine. If I am not mistaken, node exporter is started by default. One target is one routine. Also the service and node exporter process.

jakub-dzon: You're right about TSDB. node exporter is down when system metrics are disabled.

jakub-dzon: Right now there is a switch for turning the system metrics off and maybe we could add similar switch for workload metrics. When they would both be off, we could turn TSDB off.

ydayagi: ok. so when we'll do ALL of that we can do this one as well. till then no reason for this feature to be required to do more than the rest.

internal/metrics/remote_write.go
ydayagi: fixed

internal/metrics/remote_write.go
ydayagi: fixed

internal/metrics/remote_write.go
ydayagi: once PR https://github.com/project-flotta/flotta-operator/pull/148 is merged I will fix

internal/metrics/remote_write.go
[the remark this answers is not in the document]
ydayagi: fixed

internal/metrics/remote_write.go
ydayagi: fixed in previous changes

internal/metrics/remote_write_test.go
ydayagi: fixed

internal/metrics/remote_write.go
ydayagi: @pkliczewski please review my answer

jakub-dzon (approved):

masayag (approved):

pkliczewski:
:tada:

PR #117, opened 2022-03-22
Dockerfile: pull nginx image for speed the testing

eloycoto:
When testing integration, if quay is down or slow can affect our e2e
test, so in this case, I added the image in there, so no download issue
can affect the CI.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -6,12 +6,14 @@ RUN dnf copr enable project-flotta/flotta -y
 
 # Yum dependencies
 RUN dnf update -y \
-  && dnf install -y openssl procps-ng dmidecode nc \
+  && dnf install -y openssl procps-ng dmidecode nc iproute \
   yggdrasil flotta-agent-race node_exporter
 
 # Modify podman configuration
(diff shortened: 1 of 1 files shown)


tools/Dockerfile
```diff
@@ -6,13 +6,15 @@ RUN dnf copr enable project-flotta/flotta -y
...
 RUN dnf update -y \
-  && dnf install -y openssl procps-ng dmidecode nc \
+  && dnf install -y openssl procps-ng dmidecode nc iproute \
   yggdrasil flotta-agent-race node_exporter
 
 # Modify podman configuration
 RUN sed -i s/netns=\"host\"/netns=\"private\"/g /etc/containers/containers.conf && \
     sed -i s/utsns=\"host\"/utsns=\"private\"/g /etc/containers/containers.conf
 
+RUN podman pull quay.io/bitnami/nginx:1.21.6
```
eloycoto, line 16: yes, I'll change the e2e test from latest to 1.21.6

machacekondra (approved):

tools/Dockerfile
machacekondra: we use root podman, will this image be downloaded for this user?

jakub-dzon (approved):

PR #125, opened 2022-04-05
enable TLS for metrics remote writer

ydayagi:
Pass an optional CA to the device
This is the CA that signs the metrics receiver's certificate

https://issues.redhat.com/browse/ECOPROJECT-579
enable TLS in device remote write

### DIFF
--- a/internal/metrics/remote_write.go
+++ b/internal/metrics/remote_write.go
@@ -30,6 +30,7 @@ const (
 	defaultRequestRetryInterval = 10 * time.Second
 	LastWriteFileName           = "metrics-lastwrite"
 	DeviceLabel                 = "edgedeviceid"
+	ServerCAFileNamePattern     = "remote-write-ca-*.pem"
 )
 
 type RemoteWrite struct {
@@ -44,6 +45,8 @@ type RemoteWrite struct {
 	RequestRetryInterval time.Duration
 	client               WriteClient
 	isRunning            bool
+	dataDir              string
+	currentServerCaFile  string
 }
 
 func NewRemoteWrite(dataDir, deviceID string, tsdbInstance API) *RemoteWrite {
@@ -54,6 +57,7 @@ func NewRemoteWrite(dataDir, deviceID string, tsdbInstance API) *RemoteWrite {
 		RangeDuration:        defaultRequestDuration,
 		RequestRetryInterval: defaultRequestRetryInterval,
 		lastWriteFile:        path.Join(dataDir, LastWriteFileName),
+		dataDir:              dataDir,
 	}
 
 	lastWriteBytes, err := ioutil.ReadFile(newRemoteWrite.lastWriteFile)
@@ -138,33 +142,12 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
 
 	if featureDisabled {
 		r.client = nil
+		r.currentServerCaFile = ""
+		r.removeServerCaFiles()
 	} else {
-		serverURL, err := url.Parse(newConfig.URL)
+		err := r.applyConfig(newConfig)
 		if err != nil {
-			log.Errorf("metrics remote write configuration is invalid. Can not parse URL %s. Error: %s", newConfig.URL, err.Error())
 			return err
-		}
-
-		if newConfig.TimeoutSeconds == 0 {
-			return fmt.Errorf("metrics remote write configuration is invalid. TimeoutSeconds has to greater than 0")
-		}
-
-		if newConfig.RequestNumSamples == 0 {
-			return fmt.Errorf("metrics remote write configuration is invalid. RequestNumSamples has to greater than 0")
-		}
-
-		client, err := remote.NewWriteClient(r.deviceID, &remote.ClientConfig{
-			Timeout: model.Duration(newConfig.TimeoutSeconds * int64(time.Second)),
-			URL: &config_util.URL{
-				URL: serverURL,
-			},
-		})
-		if err != nil {
-			log.Error("failed creating metrics remote write client", err)
-			return err
-		}
-		r.client = &writeClient{
-			client: client,
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


ydayagi:
this PR depends on https://github.com/project-flotta/flotta-operator/pull/176. go.mod needs to be updated after the operator PR is merged

eloycoto:
shouldn't the CA be part of yggdrasil on the kickstart file?

ydayagi:
> shouldn't the CA be part of yggdrasil on the kickstart file?

kickstart is static. CA changes. feature can be disabled.

jakub-dzon (approved):

internal/metrics/remote_write.go
```diff
@@ -138,7 +142,10 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
 
 	if featureDisabled {
 		r.client = nil
+		r.currentServerCaFile = ""
+		r.removeServerCaFiles()
 	} else {
+		// parse and validate new config
```
bardielle, line 148: The update function becomes too long, please remove that "else" scope to another function

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
 			return fmt.Errorf("metrics remote write configuration is invalid. RequestNumSamples has to greater than 0")
 		}
 
+		// create TLS client config
+		clientConfig := config_util.HTTPClientConfig{}
+
+		if newConfig.CaCert != "" {
+			// create new file for CA. can't use same file cause client watches for changes in the file we pass to it.
```
gciavarrini, line 167: Maybe s/cause/bacause/ ?

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
...
+			caFile, err := ioutil.TempFile(r.dataDir, ServerCAFileNamePattern)
+			if err != nil {
+				log.Error("could not create temp file for metics remote write server CA")
```
gciavarrini, line 170: Should we log the failing filepath?

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
...
+			caFilePath := caFile.Name()
+
+			_, err = caFile.WriteString(newConfig.CaCert)
+			if err == nil {
+				err = caFile.Sync()
+				if err == nil {
+					err = caFile.Close()
+				}
+			}
+			if err != nil {
```
gciavarrini, line 183: Not sure about the `err` checks.
Does `if err != nil` refer to `_, err = caFile.WriteString(newConfig.CaCert)` ?
if yes, then this if block should be placed after `_, err = caFile.WriteString`
if no, (i.e. `err` refer to ` caFile.Close()`) then there isn't a check on `caFile.WriteString`.

Can you please clarify?

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
...
+				_ = caFile.Close()
```
gciavarrini, line 184: Is it ok to not check `err` here?

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
...
+		// create client
 		client, err := remote.NewWriteClient(r.deviceID, &remote.ClientConfig{
 			Timeout: model.Duration(newConfig.TimeoutSeconds * int64(time.Second)),
 			URL: &config_util.URL{
 				URL: serverURL,
 			},
+			HTTPClientConfig: clientConfig,
 		})
 		if err != nil {
+			_ = os.Remove(clientConfig.TLSConfig.CAFile)
```
gciavarrini, line 203: Is it ok to not check `err` here?

internal/metrics/remote_write.go
```diff
@@ -153,19 +160,58 @@ func (r *RemoteWrite) Update(config models.DeviceConfigurationMessage) error {
...
 			log.Error("failed creating metrics remote write client", err)
 			return err
 		}
+
+		_ = os.Remove(r.currentServerCaFile)
```
gciavarrini, line 208: Is it ok to not check err here?

internal/metrics/remote_write.go
```diff
@@ -384,6 +431,21 @@ func (r *RemoteWrite) writeRequest(series []Series, client WriteClient) bool {
 	return true
 }
 
+func (r *RemoteWrite) removeServerCaFiles() {
+	fileInfo, err := ioutil.ReadDir(r.dataDir)
+	if err != nil {
+		log.Errorf("could not read %s", r.dataDir)
```
gciavarrini, line 437: s/could not/cannot/ for consistency

internal/metrics/remote_write_test.go
```diff
@@ -25,21 +26,70 @@ var _ = Describe("remote write", func() {
...
 
-			remoteWrite = metrics.NewRemoteWrite("", deviceID, tsdbInstance)
+			remoteWrite = metrics.NewRemoteWrite(tmpDir, deviceID, tsdbInstance)
 		})
 
 		AfterEach(func() {
 			err := tsdbInstance.Deregister()
+			Expect(tmpDir).NotTo(BeEmpty())
+			os.RemoveAll(tmpDir)
 			Expect(err).ToNot(HaveOccurred())
```
gciavarrini, line 93: `err` in line 93 refers to line 90.
Please move 
```
   Expect(tmpDir).NotTo(BeEmpty())
   os.RemoveAll(tmpDir)
```
below it

internal/metrics/remote_write_test.go
```diff
@@ -118,6 +171,7 @@ var _ = Describe("remote write", func() {
 			Expect(remoteWrite.IsEnabled()).To(BeTrue())
 			Expect(remoteWrite.IsRunning()).To(BeTrue())
 			Expect(remoteWrite.Config).To(Equal(configMessage2.Configuration.Metrics.Receiver))
+			Expect(listCaFiles()).To(HaveLen(0))
```
gciavarrini, line 174: Is `BeEmpty` more consistent with other tests than `HaveLen(0)`?

gciavarrini (changes requested):

internal/metrics/remote_write.go
ydayagi: the function, TempFile, creates the path. So when it fails I do not have a path to log. However, I can add the dataDir. Also, I just noticed that I have a spelling error. metics ==> metrics. I don't know why my spell checker misses so much.

internal/metrics/remote_write_test.go
ydayagi: I don't think it matters. But I do believe that BeEmpty is more commonly used for strings and HaveLen for collections.

internal/metrics/remote_write.go
ydayagi: instead of checking that error occurred and then log and return, I check that err did not occur and then continue to next operation. After all operations (WriteString, Sync, Close) I check whether occurred. I then log it and return

internal/metrics/remote_write.go
ydayagi: I am already in error state. There's nothing I can do with error returned from Remove.

internal/metrics/remote_write.go
ydayagi: I do not want to fail the feature/application just because a file can not be deleted. It is a file that I am not going to use any more. During startup I delete redundant files.

internal/metrics/remote_write.go
ydayagi: done

internal/metrics/remote_write.go
ydayagi: done

internal/metrics/remote_write.go
ydayagi: done

internal/metrics/remote_write_test.go
ydayagi: done

bardielle (approved):

internal/metrics/remote_write.go
jakub-dzon: Maybe `os.RemoveAll` would be better here? It has more tolerance for unusual situations (i.e. file is missing)?

internal/metrics/remote_write.go
gciavarrini: It seems the edit is not there. Can you please double check?

internal/metrics/remote_write.go
gciavarrini: IMO this "code style" complicates a bit the understating of the code since checking if `err != nil` is the idiomatic way to determine if an error was encountered.
But if you like it more this way, that's fine.

internal/metrics/remote_write.go
ydayagi: I dont care if file is missing. I ignore the error

internal/metrics/remote_write.go
ydayagi: In this case it saves duplicity of message and code. In many cases each error has different log msg and different error handling. Here it is the same for all cases. So i save code by doing it this way. It is not a 'method' i use.  It is specifc to this case

internal/metrics/remote_write.go
jakub-dzon: Maybe you should care for the error during removal and at least log it?

ydayagi: Look at the entire block of code. We succeeded in writing a new file, we succeeded in reading the new file (prom client reads it), but we failed to delete the old file. What are the chances of that happening in the real world? 0. You want log? no problemo

pkliczewski:
@gciavarrini are you OK with this change?

gciavarrini (approved):

gciavarrini:
> @gciavarrini are you OK with this change?

yes

PR #127, opened 2022-04-08
Use custom systemd service configuration for autoupdate

machacekondra:
This commit add special systemd service configuration for the pods which
are configured to use autoupdate podman feature. The reason for this is
that podman `generate systemd` command doesnt' support generation of the
services for pod, which were created via API, or via podman play kube.
Default service generation only restarts the pod, while this custom
configuration re-create the container so the newer image could be used.

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"text/template"
 
 	"github.com/project-flotta/flotta-device-worker/internal/service"
 
@@ -21,6 +22,7 @@ import (
 	"github.com/containers/podman/v3/pkg/bindings/system"
 	"github.com/containers/podman/v3/pkg/domain/entities"
 	api2 "github.com/project-flotta/flotta-device-worker/internal/workload/api"
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkliczewski (commented):
/lgtm

internal/workload/podman/podman.go
```diff
@@ -30,13 +32,36 @@ const (
...
+[Service]
+Restart=on-failure
+RestartSec=15
+ExecStart={{ .PodmanBinary }} play kube --replace {{ .ManifestPath }}
+ExecStop={{ .PodmanBinary }} play kube --down {{ .ManifestPath }}
+Type=forking
+
+[Install]
+WantedBy=default.target
+`
```
gciavarrini, line 51: I'm wondering if a template file can be easier to modify/maintain or if there is a reason to have this inline.

internal/workload/podman/podman.go
```diff
@@ -312,16 +337,51 @@ func (p *podman) UpdateSecret(name, data string) error {
...
-	if err != nil {
-		return nil, err
+	return false
+}
+
+func (p *podman) GenerateSystemdService(workload *v1.Pod, manifestPath string, monitoringInterval uint) (service.Service, error) {
+	var svc service.Service
+	podName := workload.Name
+
+	// Since podman don't support generation of the systemd services that re-creates the pod, instead of restarting it,
```
masayag, line 354: Have you created an issue to podman for it?

masayag (approved):

jakub-dzon (approved):

internal/workload/wrapper_test.go
```diff
@@ -55,7 +55,7 @@ var _ = Describe("Workload management", func() {
 			pod := &v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "pod1"}}
 
 			newPodman.EXPECT().Run(manifestPath, authFilePath).Return([]*podman.PodReport{{Id: "id1"}}, nil)
-			newPodman.EXPECT().GenerateSystemdService(gomock.Any(), gomock.Any()).Return(svc, nil)
+			newPodman.EXPECT().GenerateSystemdService(gomock.Any(), gomock.Any(), gomock.Any()).Return(svc, nil)
```
eloycoto, line 58: shouldn't we check that at least one manifest path matches here? Mainly because maybe in future changed for some reason and we're not going to notice.

internal/workload/podman/podman.go
machacekondra: Not sure, podman uses inline code as well: https://github.com/containers/podman/blob/main/pkg/systemd/generate/common.go#L26

internal/workload/podman/podman.go
machacekondra: They seem to not support it: https://github.com/containers/podman/issues/13211
I think they don't want to create to complex code for generation, users can create their own services. I think this comment explain it well: https://github.com/containers/podman/blob/main/pkg/systemd/generate/pods.go#L293

internal/workload/podman/podman.go
```diff
@@ -312,16 +337,51 @@ func (p *podman) UpdateSecret(name, data string) error {
...
+	// for pods/containers created via the API, we must create the custom systemd service for pods, which has enabled
+	// autoupdate feature, because in case the image of the containers is updated we need to re-create the containers.
+	if !hasAutoUpdateEnabled(workload.Labels) {
```
bardielle, line 357: Maybe I missed it, but I didn't find a UT for that scenario.

pkliczewski (approved):

PR #128, opened 2022-04-08
Approval: adds Enrol process on device Worker.

eloycoto:
This integrates the enrol data message to the operator, and tries to
enrol before the device is ready to register.

If enrolment is already finished, it'll register to the device as
normal.


internal/registration/tags.go
```diff
@@ -0,0 +1,79 @@
...
+	"io/ioutil"
+	"os"
+	"reflect"
+	"strconv"
+	"time"
+
+	"github.com/pelletier/go-toml"
+)
+
+// TO BE REMOVED, we should export these functions to his own package
```
eloycoto, line 15: when this PR got merged: 
https://github.com/RedHatInsights/yggdrasil/pull/81

internal/registration/registration.go
```diff
@@ -125,8 +150,61 @@ func (r *Registration) registerDeviceWithRetries(interval int64) {
...
+
 func (r *Registration) registerDeviceOnce() error {
 
+	enrolled, err := r.enrol()
+	if err != nil {
+		return err
+	}
+
+	if !enrolled {
+		return fmt.Errorf("Device is not enrolled yet")
```
pkliczewski, line 205: Is this really an error? Do I understand correctly that it will end up here is the device is not approved yet?

internal/registration/tags.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strconv"
+	"time"
+
+	"github.com/pelletier/go-toml"
```
pkliczewski, line 13: I see there is one more [library](https://github.com/BurntSushi/toml) used in openshift. Both seems to have similar about of references. What was the reason you decided to use this one?

eloycoto: No need at all, but I wanted to use the same code as Yggdrasil, we're going to remove this when the yggdrasil code got merged.

internal/registration/registration.go
eloycoto: If we do not return an error, it'll not try to enrol again in the future.

pkliczewski: Is this visible in the logs?

eloycoto: Yes: 
https://github.com/project-flotta/flotta-device-worker/blob/dc7ddd11603141944b4a4531c29b7afea9bb6731/internal/registration/registration.go#L106-L108

pkliczewski: What do you think about no logging it if this is expected error otherwise we could log it. It would be misleading to the user.

eloycoto: at the same time is good that user knows that device is not enrolled yet (I used  for debug)

pkliczewski: I am not saying it is not needed but I am wondering whether error level is correct for this information

masayag (approved):

pkliczewski (approved):

Issue #118, opened 2022-03-27
Implement registration-retry with exponential backoff


masayag:
In current implementation, the device will try to register itself to the server every [10 seconds](https://github.com/project-flotta/flotta-device-worker/blob/main/internal/registration/registration.go#L21) if fails.

We'd like to change that behavior to retry with an exponential backoff (10s, 20s, 40s, …), that is capped at five minutes.

PR #129, opened 2022-04-13
Implement registration-retry with exponential backoff

gabriel-farache:
Pull request for Implement registration-retry with exponential backoff #118

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/opencontainers/runtime-tools v0.9.1-0.20211020193359-09d837bf40a7 // indirect
 	github.com/openshift/assisted-installer-agent v1.0.10-0.20211027185717-53b0eacfa147
 	github.com/pkg/errors v0.9.1
-	github.com/project-flotta/flotta-operator v0.0.2-0.20220404113302-b5aed0516daa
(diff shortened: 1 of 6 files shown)


eloycoto:
I'm not sure if we want this TBH, exponential backoff is ok, but a device can take too much time on that. Also, the flow will change when https://github.com/project-flotta/flotta-operator/pull/179 got merged, and I can finish the work on this: 
https://github.com/project-flotta/flotta-device-worker/pull/128 

So not sure if it something that we should spend time here.

gabriel-farache:
> I'm not sure if we want this TBH, exponential backoff is ok, but a device can take too much time on that. Also, the flow will change when [project-flotta/flotta-operator#179](https://github.com/project-flotta/flotta-operator/pull/179) got merged, and I can finish the work on this: #128
> 
> So not sure if it something that we should spend time here.

The change in the workflow that would impact the usefulness of this issue is that the device will first create enroll request, so it will be known by the system and then, the admin user can approve the enrollment. Meaning that the device is already UP so the registration should not fail much time so it won't be necessary to have exponential retries as we would not save any relevant resources.
Correct?

eloycoto:
Yup, no registration will happens at all.

PR #130, opened 2022-04-13
Rootless podman

machacekondra:
This commit add support for rootless podman.

There are multiple things needed to support rootless podman:

 0) Create a user running the podman, flotta in our case.
 1) Enable linger for the user, so systemd can run even without session.
 2) Have set the /etc/subuid /etc/subgid for the user.
 3) Create home directory for the user, to store the data and systemd
    services.
 4) Enable podman.socket for the user.
 5) Volumes directory must be owned by the user, as podman use it, not
    root.

Those prerequsites are done in spec file. For RPM based installation by
default the device-worker will execute podman as rootless. For dev-env
user will still use the rootfull pods. This could be changed the env
variables in the 'config/device-worker.toml'. We introduced two new env
variables:

  - HOME - store the home directory of the user
  - XDG_RUNTIME_DIR - store the runtime dir of the user

XDG_RUNTIME_DIR must be set for podman to properly work, this variable
is set by pam_systemd module, which is not used, as we don't switch the
user, we only use the podman socket of the user and the DBus connection,
which needs the env var as well.

The home directory is used for storing the podman data, and systemd
services.

Note the known limitation of rootless podman: https://github.com/containers/podman/blob/main/rootless.md

### DIFF
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -1,5 +1,7 @@
 %define _build_id_links none
 
+%global flotta_user flotta
+
 Name:       flotta-agent
 Version:    1.0
 Release:    1%{?dist}
@@ -15,6 +17,8 @@ Requires:       nftables
 Requires:       podman
 Requires:       yggdrasil
 
+Requires(pre): shadow-utils
+
 Provides:       %{name} = %{version}-%{release}
 Provides:       golang(%{go_import_path}) = %{version}-%{release}
 
@@ -28,9 +32,18 @@ to detect race-conditions in the e2e test
 %description
 The Flotta agent communicates with the Flotta control plane. It reports the status of the appliance and of the running PODs/containers. Agent is responsible for starting and stopping PODs that are based on commands from the control plane.
 
+%pre race
+getent group %{flotta_user} >/dev/null || groupadd %{flotta_user}; \
+getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/nologin -d /home/%{flotta_user} %{flotta_user}
+
+%pre
+getent group %{flotta_user} >/dev/null || groupadd %{flotta_user}; \
+getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/nologin -d /home/%{flotta_user} %{flotta_user}
+
 %post
-systemctl enable --now podman.socket
 systemctl enable --now nftables.service
+loginctl enable-linger %{flotta_user}
+XDG_RUNTIME_DIR="/run/user/$(getent passwd %{flotta_user} | cut -d: -f3)" su %{flotta_user} -s /bin/bash -c '/usr/bin/systemctl enable --user podman.socket'
 
 %prep
 tar fx %{SOURCE0}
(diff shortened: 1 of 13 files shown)


flotta-agent.spec
```diff
@@ -1,5 +1,8 @@
 %define _build_id_links none
 
+%global flotta_user flotta
+%global flotta_uid 10009
```
pkliczewski, line 4: why do we want to use specific uid? what happens when it is already used by different user?

tools/Dockerfile
```diff
@@ -7,13 +7,13 @@ RUN dnf copr enable project-flotta/flotta -y
...
   && dnf install -y openssl procps-ng dmidecode nc iproute \
-  yggdrasil flotta-agent-race node_exporter
+  yggdrasil flotta-agent-race node_exporter systemd-container
 
 # Modify podman configuration
 RUN sed -i s/netns=\"host\"/netns=\"private\"/g /etc/containers/containers.conf && \
     sed -i s/utsns=\"host\"/utsns=\"private\"/g /etc/containers/containers.conf
 
-RUN podman pull quay.io/bitnami/nginx:1.21.6
+#RUN podman pull quay.io/bitnami/nginx:1.21.6
```
pkliczewski, line 16: why to comment? should we remove?

flotta-agent.spec
masayag: should we add the user to the blueprint as well?

pkliczewski: I think we should

eloycoto (approved):

tools/Dockerfile
eloycoto: I guess that it's because each user will have his own images, no?

internal/service/systemd.go
```diff
@@ -133,8 +139,34 @@ func NewSystemd(name string, serviceNamePrefix string, units map[string]string)
 	return NewSystemdWithSuffix(name, serviceNamePrefix, PodSuffix, ServiceSuffix, units)
 }
 
+func newDbusConnection() (*dbus.Conn, error) {
+	if os.Getenv("XDG_RUNTIME_DIR") != "" && os.Getenv("XDG_RUNTIME_DIR") != "/run" {
```
masayag, line 143: `os.Getenv("XDG_RUNTIME_DIR")` is called 3 times in this context, can it be stored in a local variable?

masayag:
Should podman service be stopped in case there are no active connections?

jakub-dzon:
As a follow-up we need to add `EdgeWorkload` validation to make sure that host port for container is not below 1024 ([limitation of rootless podman](https://github.com/containers/podman/blob/main/rootless.md))

jakub-dzon (commented):
Will `flotta` user be allowed to control `node_exporter` service?
https://github.com/project-flotta/flotta-device-worker/blob/dc7ddd11603141944b4a4531c29b7afea9bb6731/internal/metrics/system.go#L100

flotta-agent.spec
machacekondra: We need uid > 1000, so `useradd` creates appropriate `/etc/subuid` entries. We could ask the uid to be [reserved](https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_soft_static_allocation). But I think we can uses `getent` to obtain the `uid` from `username`. So I will fix.

tools/Dockerfile
machacekondra: That's leftover, sorry

flotta-agent.spec
```diff
@@ -28,9 +32,14 @@ to detect race-conditions in the e2e test
...
 The Flotta agent communicates with the Flotta control plane. It reports the status of the appliance and of the running PODs/containers. Agent is responsible for starting and stopping PODs that are based on commands from the control plane.
 
+%pre
+getent group %{flotta_user} >/dev/null || groupadd %{flotta_user}; \
+getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -u %{flotta_uid} -s /sbin/nologin -d /home/%{flotta_user} %{flotta_user}
+
 %post
-systemctl enable --now podman.socket
 systemctl enable --now nftables.service
+loginctl enable-linger flotta
```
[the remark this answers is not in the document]
machacekondra: Right, thanks

internal/service/systemd.go
machacekondra: Yes, thanks.

machacekondra:
> Will `flotta` user be allowed to control `node_exporter` service?
> 
> https://github.com/project-flotta/flotta-device-worker/blob/dc7ddd11603141944b4a4531c29b7afea9bb6731/internal/metrics/system.go#L100

I will test it, but there should't be any issue. The only difference is that it will run as flotta user instead of root.

machacekondra:
> Should podman service be stopped in case there are no active connections?

It is. We start the `podman.socket` which triggers the `podman.service`. Podman service by default starts only for 5 seconds, and then dies. The `podman.socket` will start the service again, when there is new API request. The socket for the `podman.socket` is opened by the systemd so we don't consume any resources.

masayag:
@machacekondra pls rebase

masayag (approved):

pkliczewski (approved):

machacekondra:
Verified here: https://github.com/machacekondra/k4e-operator/runs/6177251390

PR #132, opened 2022-04-17
Bump golang version to 1.17

eloycoto:


Makefile
```diff
@@ -158,3 +165,19 @@ deploy-container-image:
...
+# go-install-tool will 'go install' any package $2 and install it to $1.
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-install-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(PROJECT_DIR)/bin go get $(2) ;\
```
masayag, line 180: shouldn't it call `go install` instead of `go get` ?

eloycoto: Go install will install it on GOPATH/bin and go get to the current /tmp folder, it's the same way as in the operator.

masayag (approved):

PR #136, opened 2022-04-26
handle gaps in TSDB for remote write

ydayagi:
it could be that the TSDB will have a large empty range.
prevent trying to read parts of the empty range until consumed entirely.

https://issues.redhat.com/browse/ECOPROJECT-722


internal/metrics/metrics.go
```diff
@@ -248,6 +250,28 @@ func (t *TSDB) MaxTime() time.Time {
...
+	blocks := t.db.Blocks()
+	blocksLen := len(blocks)
+	if blocksLen == 0 {
+		return []time.Time{}
+	}
+
+	result := make([]time.Time, blocksLen*2)
+	for i, block := range blocks {
+		result[i*2] = fromDbTime(block.MinTime())
+		result[i*2+1] = fromDbTime(block.MaxTime())
```
gciavarrini, line 269: Can a `BlocksMin` and a `BlocksMax` functions (i.e. split this function in two) make the code simpler to understand? After all, it's way way it's used in `skipEmptyRanges` [function](https://github.com/project-flotta/flotta-device-worker/blob/cb9d59bf4996acdf800f1b927b20713bcde5f0e6/internal/metrics/remote_write.go#L481). The [test](https://github.com/project-flotta/flotta-device-worker/blob/cb9d59bf4996acdf800f1b927b20713bcde5f0e6/internal/metrics/remote_write_test.go#L475) can also benefit from this change.

internal/metrics/metrics.go
```diff
@@ -248,6 +250,28 @@ func (t *TSDB) MaxTime() time.Time {
 	return fromDbTime(t.db.Head().MaxTime())
 }
 
+// HeadMinTime
+func (t *TSDB) HeadMinTime() time.Time {
+	return fromDbTime(t.db.Head().MinTime())
+}
+
+// BlockMinMax
```
eloycoto, line 258: I would add a comment on this function, not sure what is achieved here.

internal/metrics/metrics.go
```diff
@@ -248,6 +250,28 @@ func (t *TSDB) MaxTime() time.Time {
...
+func (t *TSDB) BlocksMinMax() []time.Time {
```
eloycoto, line 259: This function is 0 tested, I think that should be tested in `internal/metrics_test.go`

internal/metrics/remote_write.go
```diff
@@ -458,6 +467,45 @@ func (r *RemoteWrite) removeServerCaFiles() {
...
+			result = minTime
+		} else { // timeVal contained by this block
+			result = timeVal
+		}
+
+		break
+	}
+
+	if result.IsZero() { // no block is ahead or contains timeVal. Check head
+		minTime := r.tsdb.HeadMinTime()
```
eloycoto, line 498: Just a nit, I would remove the if/else  in there, and I' would write something simpler: 


```
 if result.IsZero() { // no block is ahead or contains timeVal. Check head
      minTime := r.tsdb.HeadMinTime()
      if timeVal.Sub(minTime) >= 0 { 
        return timeVal
      }
}
return result
```

eloycoto (changes requested):
Just some code that needs to be tested, and a break in a loop that I cannot fully understand. 


Also, now that you're touching that function, if you can clean this error check, it'll be great: 
https://github.com/project-flotta/flotta-device-worker/blob/5a76e714a9abe60ac84eedf7d5a88d3f76e63a05/internal/metrics/remote_write.go#L308-L315

Thanks.

internal/metrics/metrics.go
ydayagi: I did not understand. I call 'block.MinTime'. So how is BlockMin different? can u suggest code edit? and.....does it really matter? cause this change is complex anyhow and requires a lot of attention.

internal/metrics/metrics.go
ydayagi: it is a very small and simple one. It's not worth the effort. Actually, there are others such as AddVector that were not tested. What test cases would u suggest?

internal/metrics/remote_write.go
```diff
@@ -229,9 +229,11 @@ func (r *RemoteWrite) applyConfig(newConfig *models.MetricsReceiverConfiguration
 		return err
 	}
 
-	err = os.Remove(r.currentServerCaFile)
-	if err != nil {
-		log.Errorf("failed removing file %s with error %s", r.currentServerCaFile, err.Error())
+	if r.currentServerCaFile != "" {
+		err = os.Remove(r.currentServerCaFile)
+		if err != nil {
+			log.Errorf("failed removing file %s with error %s", r.currentServerCaFile, err.Error())
```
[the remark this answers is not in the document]
ydayagi: that is not quite accurate. %v is for printing the struct/object. Here we just want a string message. We've been doing it all over the code.

internal/metrics/metrics.go
```diff
@@ -248,6 +250,28 @@ func (t *TSDB) MaxTime() time.Time {
...
+	blocks := t.db.Blocks()
+	blocksLen := len(blocks)
+	if blocksLen == 0 {
+		return []time.Time{}
+	}
+
+	result := make([]time.Time, blocksLen*2)
+	for i, block := range blocks {
+		result[i*2] = fromDbTime(block.MinTime())
```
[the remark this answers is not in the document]
ydayagi: It is very clear what this piece of code does. The slice is made of pairs so i*2 and i*2+1 are what u need for accessing the pairs. it is more efficient than append.

internal/metrics/remote_write.go
```diff
@@ -458,6 +467,45 @@ func (r *RemoteWrite) removeServerCaFiles() {
...
+			continue
+		}
+
+		if timeVal.Sub(minTime) < 0 { // timeVal precedes this block
+			result = minTime
+		} else { // timeVal contained by this block
+			result = timeVal
+		}
+
+		break
```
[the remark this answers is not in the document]
ydayagi: first part of the loop skips blocks by calling continue.
The other part (this one) breaks cause it found the desired block.

internal/metrics/remote_write.go
ydayagi: you mean
`resukt = r.tsdb.HeadMinTime()
`
and just '>' for the if.
OK

internal/metrics/metrics.go
gciavarrini: I'm sorry If my comment wasn't clear. 

You're using a slice like that contains `[block1_min, block1_max, block2_min, block2_max, ... ]` thus you need to iterate over the slice with`results[i*2]` and `results[i*2+1]`.
It isn't straightforward to understand the sense of the code, in my opinion.

I'm suggesting to use two different functions to retrieve two different slices (one for the max values and one for the min values) like these:

`func (t *TSDB) BlocksMin() []time.Time`
`func (t *TSDB) BlocksMax() []time.Time`

In this way, you can iterate over the slices in a "standard" way, the function goals will be clear and, at the same, time we will make life easier for a new contributor  :)

ydayagi: In your example, you are referring to reading the min/max. But in my code I do not read from the slice. I create the slice. so, these functions will not help.
To me, this is very simple arithmetic that for us should be a no-brainer. I am actually, quite surprised that it is making so much noise. This is a very (very) common pattern in many libraries and functions. I can easily change my code based on everyone's comments. I just don't think it's the right way to perform our PR reviews. Think of the number of functions we add. I think it makes the code more difficult to track. Your call.

gciavarrini: I just gave you my opinion, feel free to make your own decisions

internal/metrics/metrics.go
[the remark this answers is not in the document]
ydayagi: @eloycoto look at my suggestion in my repsonse to Gloria. I think it solves the whole arithmetic issue.

internal/metrics/metrics.go
ydayagi: @gciavarrini fixed

internal/metrics/metrics.go
ydayagi: fixed

internal/metrics/metrics.go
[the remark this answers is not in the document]
ydayagi: fixed. I also changed to append.

internal/metrics/remote_write.go
ydayagi: done

jakub-dzon (approved):

internal/metrics/remote_write.go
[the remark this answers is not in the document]
masayag: the recommended way is %w or %v. in most cases %v and %s will return the same result, except for nil. but in this case, err shouldn't be nil.

masayag (approved):

PR #138, opened 2022-04-27
Use custom XDG_RUNTIME_DIR

machacekondra:
This patch add custom XDG_RUNTIME_DIR so user can customize it as
needed. For RPM installation we use flotta user XDG_RUNTIME_DIR env var
unless user overwrite it.


arielireni:
LGTM

masayag (approved):

PR #142, opened 2022-05-06
Build test container image for each PR

machacekondra:
This PR add a workflow job, that build the test container for each PR.

Every container will be stored as zipped artifact. You can later use this artifact to run manual workflow job in `flotta-operator` repo to test the your code with our e2e tests.

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -1,0 +1,54 @@
+---
+name: Container build
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  rpm:
+    name: Build rpm
+    container: fedora:latest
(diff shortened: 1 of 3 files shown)


.github/workflows/container.yaml
```diff
@@ -0,0 +1,54 @@
...
+          path: tools/
+
+      - name: Build the container image
+        run: |
+          docker build --build-arg quay_expiration=1d --build-arg flotta_agent_rpm=flotta-agent-race*.rpm ./tools/ -t edgedevice
+
+      - name: Upload image
+        uses: ishworkh/docker-image-artifact-upload@v1
+        with:
+          image: edgedevice
```
masayag, line 54: to which repo will this be uploaded, and couldn't it cause to an increased storage consumption?

PR #143, opened 2022-05-06
Implement deltas of heartbeat information

gabriel-farache:
This PR introduce 2 changes in the heartbeat data when `spec.heartbeat.hardwareProfile.include` is set to true:

1. After 1st heartbeat, only mutable hardware information are sent in the heartbeat data

- The data sent in the heartbeat packet are: CPU, SystemVendor, Interfaces and Hostname.
- Mutable information are the information that can change during the device's lifecycle. Currently, amongst data sent in the heartbeat packet, there are 2 inputs that are mutables: Hostname and Interfaces.

2. The parameter `spec.heartbeat.hardwareProfile.scope` is taken into account
- If `scope==full` all the mutable information are sent in the hardware parameter of the heartbeat.
- If `scope==delta`, the only the mutable information that changed since last heartbeat will be included in the heartbeat.

Signed-off-by: Gabriel Farache gfarache@redhat.com


pkliczewski (commented):
/lgtm

internal/hardware/hardware.go
```diff
@@ -36,6 +62,22 @@ func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
 		}
 		hardwareInfo.Interfaces = append(hardwareInfo.Interfaces, newInterface)
 	}
+}
 
-	return &hardwareInfo, nil
+func (s *Hardware) Init() {
+	if s.Dependencies == nil {
```
eloycoto, line 68: I would take xtra care here, this will be called from heartbeat(goroutine) and registration (can be another goroutine) some race conditions can happen here. Atomic or mutex is recommended.

internal/heartbeat/heartbeat.go
```diff
@@ -85,6 +90,42 @@ func (s *HeartbeatData) RetrieveInfo() models.Heartbeat {
 	return heartbeatInfo
 }
 
+func (s *HeartbeatData) buildHardwareInfo() *models.HardwareInfo {
+	var hardwareInfo *models.HardwareInfo
+	var err error
+	if s.previousMutableHardwareInfo == nil {
+		// Only send all Hw info (mutable + immutable) for the 1st heartbeat, then send only mutable hw info
```
eloycoto, line 97: Umm, what happens if heartbeat got 401/403 and there is no data submitted to the API, shouldn't this be on the send() function if everything is ok?

internal/heartbeat/heartbeat.go
```diff
@@ -85,6 +90,42 @@ func (s *HeartbeatData) RetrieveInfo() models.Heartbeat {
...
+		hardwareInfo, err = s.hardware.GetHardwareInformation()
+		if err != nil {
+			log.Errorf("cannot get full hardware information. DeviceID: %s; err: %v", s.workloadManager.GetDeviceID(), err)
+		}
+		s.previousMutableHardwareInfo = s.hardware.CreateHardwareMutableInformation()
+		if err != nil {
+			log.Errorf("cannot get mutable hardware information. DeviceID: %s; err: %v", s.workloadManager.GetDeviceID(), err)
+		}
```
eloycoto, line 105: as a rule of dumb, I will do a `return hardwareInfo` here, and remove the else, it's a bit clear what it's doing ;-)

internal/heartbeat/heartbeat.go
```diff
@@ -85,6 +90,42 @@ func (s *HeartbeatData) RetrieveInfo() models.Heartbeat {
...
+
+func (s *HeartbeatData) getMutableHardwareInfoDelta() *models.HardwareInfo {
+	hardwareMutableInfo := s.hardware.CreateHardwareMutableInformation()
+	var hardwareInfo *models.HardwareInfo
+
+	if s.configManager.GetDeviceConfiguration().Heartbeat.HardwareProfile.Scope == SCOPE_DELTA {
+		log.Debugf("Checking if mutable hardware information change between heartbeat (scope = delta). DeviceID: %s", s.workloadManager.GetDeviceID())
+		hardwareInfo = hw.GetMutableHardwareInfoDelta(*s.previousMutableHardwareInfo, *hardwareMutableInfo)
+		s.previousMutableHardwareInfo = hardwareMutableInfo
+	} else {
```
eloycoto, line 121: same here, I would remove this else.

internal/heartbeat/heartbeat_test.go
```diff
@@ -152,6 +158,165 @@ var _ = Describe("Heartbeat", func() {
 			Expect(heartbeatInfo.Status).To(Equal("up"))
 			Expect(heartbeatInfo.Workloads).To(HaveLen(0))
 		})
+
+		It("Report workload delta enable", func() {
+			//given
+			interfaceMock := NewFilledInterfaceMock(1500, "eth0", "f8:75:a4:a4:00:fe", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.18/24", "fe80::d832:8def:dd51:3527/128", "de90::d832:8def:dd51:3527/128"}, true, false, true, 1000)
+			depMock := &util.MockIDependencies{}
```
eloycoto, line 165: I guess that gomock is not a option here, no? I want to avoid to add another dependency if we can!

internal/hardware/hardware.go
```diff
@@ -1,30 +1,56 @@
...
 func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
-	d := util.NewDependencies("/")
-	cpu := inventory.GetCPU(d)
-	hostname := inventory.GetHostname(d)
-	systemVendor := inventory.GetVendor(d)
-	interfaces := inventory.GetInterfaces(d)
-
+	s.Init()
 	hardwareInfo := models.HardwareInfo{}
+	err := s.GetHardwareImmutableInformation(&hardwareInfo)
```
eloycoto, line 18: We got all these tested on internal/heartbeat, maybe should we move the test to this package?

eloycoto (changes requested):
If you can cherry-pick the go.mod with the main commit will be great. At the same hand, there is a merge commit that need to be deleted. 

Some comments in the PR, but overall, great work!

internal/hardware/hardware.go
gabriel-farache: You mean that the function calling Init could be called simultaneously by different goroutines? 
Because, actually, this Init may be private and should only be called by internal function, I missed this one
I added the dependencies field mainly to allow testing the retrieval of hardware info 
IMO, using mutex could consume too much resource compared to the service provided

internal/heartbeat/heartbeat.go
gabriel-farache: good idea for clarity

internal/heartbeat/heartbeat.go
gabriel-farache: you would return in the if?
I like to keep a single return in my function, except when an error is found and that the workflow should be interrupted but it's not the case here

internal/heartbeat/heartbeat_test.go
gabriel-farache: TBH I used what was done in the assisted-intaller-agent project for mocks
WOuld gomock be usable with mocking library calls?

internal/heartbeat/heartbeat.go
gabriel-farache: What do you mean?

internal/hardware/hardware.go
gabriel-farache: You mean to add test for hardware to do unit test of GetHardwareInformation and GetHardwareImmutableInformation?

eloycoto: yeah, and try to keep heartbeat hardware agnostic.

internal/heartbeat/heartbeat_test.go
eloycoto: I'm not sure :D just asking, it'll be good if we keep consistent with tools ;-)

internal/heartbeat/heartbeat.go
eloycoto: I would refactor like this: 

```
func (s *HeartbeatData) getMutableHardwareInfoDelta() *models.HardwareInfo {
	hardwareMutableInfo := s.hardware.CreateHardwareMutableInformation()
	hardwareInfo := hardwareMutableInfo

	if s.configManager.GetDeviceConfiguration().Heartbeat.HardwareProfile.Scope == SCOPE_DELTA {
		log.Debugf("Checking if mutable hardware information change between heartbeat (scope = delta). DeviceID: %s", s.workloadManager.GetDeviceID())
		hardwareInfo = hw.GetMutableHardwareInfoDelta(*s.previousMutableHardwareInfo, *hardwareMutableInfo)
	}
	s.previousMutableHardwareInfo = hardwareMutableInfo
	return hardwareInfo
}
```

internal/heartbeat/heartbeat.go
eloycoto: you send the first Hearbeat(Full info) and operator returns 401. 
the next thing that you're going to send, it's a delta, operator will know about the full hardware?

internal/heartbeat/heartbeat.go
gabriel-farache: Ah! Yes I see, it could even be 
```
func (s *HeartbeatData) getMutableHardwareInfoDelta() *models.HardwareInfo {
	hardwareInfo := s.hardware.CreateHardwareMutableInformation()

	if s.configManager.GetDeviceConfiguration().Heartbeat.HardwareProfile.Scope == SCOPE_DELTA {
		log.Debugf("Checking if mutable hardware information change between heartbeat (scope = delta). DeviceID: %s", s.workloadManager.GetDeviceID())
		hardwareInfo = hw.GetMutableHardwareInfoDelta(*s.previousMutableHardwareInfo, *hardwareInfo)
	}
	s.previousMutableHardwareInfo = hardwareMutableInfo
	return hardwareInfo
}
```

internal/heartbeat/heartbeat.go
gabriel-farache: Indeed, I should add something to reset previous when handling error [here](https://github.com/project-flotta/flotta-device-worker/blob/c8ae0c9b7c8fb78f810514c3c8e30c2cf9eccf2f/internal/heartbeat/heartbeat.go#L160)   and [here](https://github.com/project-flotta/flotta-device-worker/blob/c8ae0c9b7c8fb78f810514c3c8e30c2cf9eccf2f/internal/heartbeat/heartbeat.go#L171)

And I should add one [here](https://github.com/project-flotta/flotta-device-worker/blob/c8ae0c9b7c8fb78f810514c3c8e30c2cf9eccf2f/internal/heartbeat/heartbeat.go#L176)

So I'll have to introduce a bool to reset previous only if the first send fails

internal/heartbeat/heartbeat.go
```diff
@@ -65,10 +73,7 @@ func (s *HeartbeatData) RetrieveInfo() models.Heartbeat {
 	config := s.configManager.GetDeviceConfiguration()
 	var hardwareInfo *models.HardwareInfo
 	if config.Heartbeat.HardwareProfile.Include {
-		hardwareInfo, err = s.hardware.GetHardwareInformation()
-		if err != nil {
-			log.Errorf("cannot get hardware information. DeviceID: %s; err: %v", s.workloadManager.GetDeviceID(), err)
-		}
+		hardwareInfo = s.buildHardwareInfo()
```
masayag, line 76: @jakub-dzon why should the flow proceed if there was an error to collect HW info?

internal/heartbeat/heartbeat.go
```diff
@@ -85,6 +90,42 @@ func (s *HeartbeatData) RetrieveInfo() models.Heartbeat {
...
+func (s *HeartbeatData) getMutableHardwareInfoDelta() *models.HardwareInfo {
+	hardwareMutableInfo := s.hardware.CreateHardwareMutableInformation()
+	var hardwareInfo *models.HardwareInfo
+
+	if s.configManager.GetDeviceConfiguration().Heartbeat.HardwareProfile.Scope == SCOPE_DELTA {
+		log.Debugf("Checking if mutable hardware information change between heartbeat (scope = delta). DeviceID: %s", s.workloadManager.GetDeviceID())
+		hardwareInfo = hw.GetMutableHardwareInfoDelta(*s.previousMutableHardwareInfo, *hardwareMutableInfo)
+		s.previousMutableHardwareInfo = hardwareMutableInfo
+	} else {
+		hardwareInfo = hardwareMutableInfo
```
masayag, line 122: In the resolved conversation it was mentioned the `else` block can be removed and pull shared logic after the `if` statement

gabriel-farache: I did not pushed the changes yet :)

internal/heartbeat/heartbeat.go
gabriel-farache: IMO HW info are less important than receiving a heartbeat as if we do not send the heartbeat, there can be other implications at the other end of the tunnel (like disable workload or whatever) whereas the device is still UP
Nevertheless, if we let the flow proceed if a minor error occurs, we should notify it in the heartbeat

internal/heartbeat/heartbeat.go
```diff
@@ -23,29 +23,37 @@ import (
 	pb "github.com/redhatinsights/yggdrasil/protocol"
 )
 
+const (
+	SCOPE_DELTA = "delta"
```
masayag, line 27: naming convention is `ScopeDelta` or `scopeDelta` (if not exported)

internal/hardware/hardware.go
```diff
@@ -36,6 +70,26 @@ func (s *Hardware) GetHardwareInformation() (*models.HardwareInfo, error) {
...
+}
 
-	return &hardwareInfo, nil
+func (s *HardwareInfo) init() {
+	if s.Dependencies == nil {
+		s.Dependencies = util.NewDependencies("/")
+	}
+}
+
+func (s *HardwareInfo) GetMutableHardwareInfoDelta(hardwareMutableInfoSource models.HardwareInfo, hardwareMutableInfoTarget models.HardwareInfo) *models.HardwareInfo {
```
masayag, line 81: in terms of naming, the relation between the parameters isn't a source and a target.
it is prev/new or former/updated.

gabriel-farache: Ok, if you feel that it would be more understandable, I'll do the change
FYI, I named target and source as it's checking the source variable against the target variable and only the values of the target that are different will be kept.

masayag: I don't have a strong opinion about it. it can remain as is.

internal/heartbeat/heartbeat.go
gabriel-farache: I added check [here](https://github.com/project-flotta/flotta-device-worker/blob/c8ae0c9b7c8fb78f810514c3c8e30c2cf9eccf2f/internal/heartbeat/heartbeat.go#L235) and I reset the previous if there is an error during the 1st heartbeat

internal/hardware/hardware.go
gabriel-farache: done

masayag (approved):

eloycoto:
Overall PR looks good, some minor style details :-)

internal/heartbeat/heartbeat_test.go
```diff
@@ -152,6 +156,180 @@ var _ = Describe("Heartbeat", func() {
...
+					IPV4Addresses: []string{"127.0.0.1", "0.0.0.0"},
+				}},
+			}).Times(2)
+			heartbeatInfo = hbData.RetrieveInfo()
+			Expect(heartbeatInfo.Hardware.CPU).To(BeNil())
+			Expect(heartbeatInfo.Hardware.Hostname).To(Equal("localhostNEW"))
+			Expect(heartbeatInfo.Hardware.Interfaces).To(BeNil())
+			Expect(heartbeatInfo.Hardware.SystemVendor).To(BeNil())
+
+			// Hardware checks delta time interface	 change
```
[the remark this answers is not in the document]
gabriel-farache: I'll check what a table driven test is to see how I could apply it here but I'll definitely split the test

internal/hardware/hardware_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	AfterEach(func() {
+
+	})
+
+	Context("Hardware info test", func() {
+
+		It("Hw full info", func() {
+			hwInfo, err := hw.GetHardwareInformation()
+			Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
gabriel-farache: ok I'll add the comments but // given and // when will be merge inside a single comment as all the preps are done in the before

internal/hardware/hardware_test.go
```diff
@@ -0,0 +1,206 @@
...
+			hwInfo := hw.CreateHardwareMutableInformation()
+
+			Expect(hwInfo.CPU).To(BeNil())
+			Expect(hwInfo.Hostname).To(Equal("localhost"))
+			Expect(hwInfo.Interfaces).To(Not(BeNil()))
+			Expect(hwInfo.SystemVendor).To(BeNil())
+
+		})
+
+		It("Hw info delta", func() {
```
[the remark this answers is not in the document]
gabriel-farache: I hesitated of splitting it but I wanted to keep the workflow so I'll add the expected behaviour in the // check comments to make it clearer and if you find it's not clear enough I'll split it

internal/heartbeat/heartbeat.go
```diff
@@ -191,7 +226,15 @@ func (s *Heartbeat) pushInformation() error {
 		Content:   content,
 		Directive: "heartbeat",
 	}
-	return s.send(data)
+
+	err = s.send(data)
+
+	if err == nil {
+		s.firstHearbeat = false
```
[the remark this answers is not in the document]
gabriel-farache: Indeed it's not tested, actually there is no real tests on the behaviour of anything inside the pushInformation; there are only tests checking whether the ticker is started or not, there are not checks about the things that are done in the ticker goroutine
I'll see if with another mocked Dispatcher (one which will return an error when sending and that will keep an history of the data) associated with a sleep I can test that 2 successives send are having similar data for the hardware part. I'm not very familiar with go so if you think of a "go way" to do it, I'll gladfully take the advice

internal/hardware/hardware_test.go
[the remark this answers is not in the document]
gabriel-farache: I split it

eloycoto (approved):
:tada:  congrats @gabriel-farache awesome work!

PR #146, opened 2022-05-10
Makefile: Fix issues with paths

eloycoto:
When installing device-worker we should follow the same way as
YGGDrasild, and using /var/ is not a option at all because YGGdrasild is
not using like that:

https://github.com/RedHatInsights/yggdrasil/blob/cdde836c519ac72302714b5f7ae90e45de7c79a7/Makefile#L29-L39

On RPM, because prefix is set, there is no problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,10 @@ BUILDROOT ?=
 DOCKER ?= podman
 IMG ?= quay.io/project-flotta/edgedevice:latest
 
-ifeq ($(OS),fedora)
-	LIBEXECDIR ?= /var/local/libexec
-	SYSCONFDIR ?= /var/local/etc
-else
-	LIBEXECDIR ?= /var/libexec
-	SYSCONFDIR ?= /etc
-endif
+# Installation directories
+PREFIX        ?= /usr/local
+LIBEXECDIR    ?= $(PREFIX)/libexec
+SYSCONFDIR    ?= $(PREFIX)/etc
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,7 +130,8 @@ install-arm64: build-arm64
 	sudo install -D -m 755 ./bin/device-worker-aarch64 $(LIBEXECDIR)/yggdrasil/device-worker
 
 uninstall: clean
-	sudo rm -rf $(SYSCONFDIR)/yggdrasil/device/*
+	rm -rf $(SYSCONFDIR)/yggdrasil/device/*
```
masayag, line 133: sudo? it will not be supported as a user without access write to those folders.

eloycoto: I mean, we shouldn't use sudo on makefiles at all.To uninstall you can should type `sudo make uninstall` IMHO.

masayag: in that case we'd be better with getting rid of all the `sudo`, document the need to run specific rules with `sudo` and enforce from within the the makefile having running as root, e.g. by:

```
ifneq ($(shell id -u), 0)
    @echo "You must be root to perform this action."
else
    rm -rf $(SYSCONFDIR)/yggdrasil/device/*
    rm -rf $(LIBEXECDIR)/yggdrasil/device-worker
endif
```

WDYT?

masayag (approved):

PR #156, opened 2022-05-23
Check heartbeat response before trying to parse it

gabriel-farache:
This PR fixes parsing error, that has no impact but produces misleading log, detected by @pkliczewski 


gabriel-farache:
(!) PR #155 shall be merged before merging this one: I needed a feature implemented by PR #155 to test my change

pkliczewski (commented):
/lgtm

ydayagi:
@gabriel-farache this PR should fix https://issues.redhat.com/browse/ECOPROJECT-817. Please take a look and update.

internal/heartbeat/heartbeat.go
```diff
@@ -124,12 +139,16 @@ func (s *HeartbeatData) getMutableHardwareInfoDelta(currentMutableHwInfo models.
...
-	ticker           *time.Ticker
-	dispatcherClient pb.DispatcherClient
-	data             *HeartbeatData
-	lock             sync.Mutex
-	reg              *registration.Registration
-	firstHearbeat    bool
+	ticker                *time.Ticker
+	dispatcherClient      pb.DispatcherClient
+	data                  *HeartbeatData
+	lock                  sync.Mutex
```
tupyy, line 145: my 2 cents. 
You could compose your struct with `sync.Mutex` like
```
type Heartbeat struct {
   sync.Mutext
}
```
and lock it with `s.Lock()`. IMO it is cleaner than `s.lock.Lock()`.

internal/heartbeat/heartbeat.go
```diff
@@ -146,21 +165,32 @@ func NewHeartbeatService(dispatcherClient pb.DispatcherClient, configManager *cf
...
+
 func (s *Heartbeat) send(data *pb.Data) error {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
-
+	log.Debugf("Heartbeat send: Sending data: %+v; Device ID: %s", data, s.data.workloadManager.GetDeviceID())
 	response, err := s.dispatcherClient.Send(ctx, data)
+	log.Debugf("Heartbeat send: Response: %+v, err: %s; Device ID: %s", response, err, s.data.workloadManager.GetDeviceID())
```
tupyy, line 185: `%+v` for errors

internal/heartbeat/heartbeat.go
```diff
@@ -173,9 +203,14 @@ func (s *Heartbeat) send(data *pb.Data) error {
 	s.reg.RegisterDevice()
 	// Sending again the heartbeat info with the right info.
 	_, err = s.dispatcherClient.Send(ctx, data)
+
 	return err
 }
 
+func isResponseEmpty(response *pb.Response) bool {
+	return len(response.Response) <= 0
```
tupyy, line 211: why the `<`?. To be honest `len(response.Response) == 0` is clear enough. IMO you don't need a function here

internal/heartbeat/heartbeat.go
```diff
@@ -195,12 +230,19 @@ func (s *Heartbeat) Init(config models.DeviceConfigurationMessage) error {
 }
 
 func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) error {
+	s.tickerLock.Lock()
```
tupyy, line 233: why not using defer to unlock it?

internal/heartbeat/heartbeat.go
```diff
@@ -231,35 +274,45 @@ func (s *Heartbeat) pushInformation() error {
...
+	log.Debugf("pushInformation: sending content %+v; DeviceID: %s;", content, deviceId)
 	err = s.send(data)
 
+	s.pushInfoLock.Lock()
+	defer s.pushInfoLock.Unlock()
 	if err == nil {
 		s.firstHearbeat = false
 	} else if s.firstHearbeat {
-		s.data.previousMutableHardwareInfo = nil
+		s.data.SetPreviousHardwareInfo(nil)
```
tupyy, line 285: hmm. passing `nil`. IMO it was better before

jakub-dzon:
Pleae use RWMutex, like mentioned in #160

internal/heartbeat/heartbeat.go
```diff
@@ -195,12 +230,19 @@ func (s *Heartbeat) Init(config models.DeviceConfigurationMessage) error {
...
 
 func (s *Heartbeat) Update(config models.DeviceConfigurationMessage) error {
+	s.tickerLock.Lock()
 	periodSeconds := s.getInterval(*config.Configuration)
-	log.Infof("reconfiguring ticker with interval: %v. DeviceID: %s", periodSeconds, s.data.workloadManager.GetDeviceID())
-	if s.ticker != nil {
-		s.ticker.Stop()
+	if s.previousPeriodSeconds <= 0 || s.previousPeriodSeconds != periodSeconds {
+		s.log.Debugf("Heartbeat configuration update: periodSeconds changed from %d to %d; Device ID: %s", s.previousPeriodSeconds, periodSeconds, s.data.workloadManager.GetDeviceID())
+		s.tickerLock.Unlock()
```
eloycoto, line 237: maybe do defer on the line 233 is the best option tbh.

internal/heartbeat/heartbeat.go
eloycoto: BTW: I'm a bit against this one, having a lock like this means that someone from an external package can lock the struct and deadlocks can happen without notice.

I'm ok to have it, but I remember to debug a hard deadlock because of a lock from another package.

internal/heartbeat/heartbeat.go
gabriel-farache: good point for `nil` testing response 
And I want to keep it in a function in case the way we test for empty response changes: at first, I converted the response into a JSON string and I tested if the JSON string was `{}`, but then I found a better way. So if it become more complicated, at least, we only have to update the function and that will have no impact on the overall readability

internal/heartbeat/heartbeat.go
gabriel-farache: I have to lock it to avoid having race with `s.previousPeriodSeconds`
This will be a ReadLock when I'll be done implementing the change using RWMutex

internal/heartbeat/heartbeat.go
gabriel-farache: Had to unlock here to avoid deadlock when call s.initTicker

internal/heartbeat/heartbeat.go
gabriel-farache: Yep I have to protect it, if I don't there may be a data race as the `pushInformation` method is called be the ticker goroutine so if for any reason, an new `initTicker` were called while a tick occurs, we could have simultaneous access
If the send fails, I have to reset the previous state if it was the 1st call otherwise, only delta will be sent in the heartbeat while the client never received the full first heartbeat as it should expect

gabriel-farache:
> @gabriel-farache this PR should fix https://issues.redhat.com/browse/ECOPROJECT-817. Please take a look and update.

It's more https://github.com/project-flotta/flotta-device-worker/pull/155 that should fix this issue IMO, as it's this PR that implements the "check the configuration for changes before updating" feature

internal/heartbeat/heartbeat.go
gabriel-farache: yeah mutex shall be private, it will be a nightmare to debug in case of issue otherwise

internal/heartbeat/heartbeat.go
```diff
@@ -262,16 +275,21 @@ func (s *Heartbeat) pushInformation() error {
...
 		Directive: "heartbeat",
 	}
-
+	log.Debugf("pushInformation: sending content %+v; DeviceID: %s;", content, deviceId)
 	err = s.send(data)
 
-	s.pushInfoLock.Lock()
-	defer s.pushInfoLock.Unlock()
 	if err == nil {
+		s.pushInfoLock.Lock()
```
tupyy, line 282: WDYT about this:
```go
if err = s.send(data); err != nil {
   s.pushInfoLock.RLock()
   defer s.pushInfoLock.RUnlock()
   if s.firstHearbeat {
	s.data.SetPreviousHardwareInfo(nil)
   }
   log.Debugf("pushInformation: sending content results %s; DeviceID: %s;", err, deviceId)
   return err
}


s.pushInfoLock.Lock()
defer s.pushInfoLock.Unlock()
s.firstHearbeat = false
return nil
```

gabriel-farache: Ok, I understand now what you meant by "It's more go style" => test `err` for `!=nil` and return the error instead of testing for `==nil`
I'll do that

internal/heartbeat/heartbeat.go
```diff
@@ -219,19 +230,20 @@ func (s *Heartbeat) Init(config models.DeviceConfigurationMessage) error {
...
 	if s.previousPeriodSeconds <= 0 || s.previousPeriodSeconds != periodSeconds {
 		s.log.Debugf("Heartbeat configuration update: periodSeconds changed from %d to %d; Device ID: %s", s.previousPeriodSeconds, periodSeconds, s.data.workloadManager.GetDeviceID())
-		s.tickerLock.Unlock()
+		// Had to unlock here to avoid deadlock when call s.initTicker
+		s.tickerLock.RUnlock()
 		s.log.Infof("reconfiguring ticker with interval: %v. DeviceID: %s", periodSeconds, s.data.workloadManager.GetDeviceID())
 		if s.ticker != nil {
 			s.ticker.Stop()
 		}
 		s.initTicker(periodSeconds)
```
eloycoto, line 243: maybe here, the lock should happens inside the initticker and have a function called stopTicker that does lock/unlock for simplicity

internal/heartbeat/heartbeat_test.go
```diff
@@ -644,6 +662,42 @@ func (d *Dispatcher) GetConfig(ctx context.Context, in *pb.Empty, opts ...grpc.C
 	return nil, nil
 }
 
+type DispatcherEmptyResponse struct {
+	latestData *pb.Data
+	hwInfoList []*models.HardwareInfo
+	mu         sync.Mutex
```
eloycoto, line 668: if we use lock across the project, I'll keep here lock instead of mu

internal/heartbeat/heartbeat.go
eloycoto: :+1:

pkliczewski (approved):

internal/heartbeat/heartbeat.go
```diff
@@ -262,24 +277,29 @@ func (s *Heartbeat) pushInformation() error {
...
-
+	log.Debugf("pushInformation: sending content %+v; DeviceID: %s;", content, deviceId)
 	err = s.send(data)
+	log.Debugf("pushInformation: sending content results %s; DeviceID: %s;", err, deviceId)
 
+	if err != nil {
+		s.pushInfoLock.RLock()
+		defer s.pushInfoLock.RUnlock()
+		if s.firstHearbeat {
+			s.data.SetPreviousHardwareInfo(nil)
```
ydayagi, line 288: isn't it already nil during the first heartbeat?

ydayagi:
I think it is better to use 'commit amend' for each change instead of multiple commits. It is difficult to diff final result versus original code.

internal/heartbeat/heartbeat.go
gabriel-farache: No, the variable is set in `s.data.RetrieveInfo()` and the send is performed after; so if the send fails, we have to be reset the variable value to `nil` to ensure the next call will gather all the information and not just the delta depending on the `Scope` that is set.

internal/heartbeat/heartbeat.go
```diff
@@ -295,11 +315,14 @@ func (s *Heartbeat) initTicker(periodSeconds int64) {
 
 func (s *Heartbeat) Deregister() error {
 	s.log.Infof("stopping heartbeat ticker. DeviceID: %s", s.data.workloadManager.GetDeviceID())
-	if s.ticker != nil {
+	s.stopTicker()
+	return nil
+}
+
+func (s *Heartbeat) stopTicker() {
```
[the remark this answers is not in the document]
gabriel-farache: Well not really; `HasStarted` only checks if the ticker is created and `stopTicker` is stoping the ticker so it needs to check that `ticker != nil` before stoping it, that's why I used `HasStarted` here. But I agree I could simply check for the `nil` as it's the only thing the matters here.
`HasStarted` could become more complex if we want to check other things other than the `nil` value so I'll keep the function.
And, for readability, it's better to separate `stopTicker` from `Deregister`, as again, deregister could become more complex than just stoping the ticker. This way it will be easier to make new changes to it

internal/heartbeat/heartbeat.go
```diff
@@ -37,7 +39,7 @@ type HeartbeatData struct {
 	hardware                        hw.Hardware
 	osInfo                          *os2.OS
 	previousMutableHardwareInfo     *models.HardwareInfo
-	previousMutableHardwareInfoLock sync.Mutex
+	previousMutableHardwareInfoLock sync.RWMutex
```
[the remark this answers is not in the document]
gabriel-farache: WDYM? `previousMutableHardwareInfo` has to be protected to avoid data races when read/write accesses
Or do you mean I should use sync/atomic [LoadPointer](https://pkg.go.dev/sync/atomic#LoadPointer) and [StorePointer](https://pkg.go.dev/sync/atomic#StorePointer) function instead?

internal/heartbeat/heartbeat.go
gabriel-farache: I used atomic to avoid having to `RLock` then manually (meaning no `defer`) ` when checking the value of `previousPeriodSeconds`  then RUnlock`, `Lock` and manually `Unlock` when setting it (I have to be inside the condition) 
Using the atomic allows me avoid all this

gabriel-farache:
> I think it is better to use 'commit amend' for each change instead of multiple commits. It is difficult to diff final result versus original code.

I will rebase after the reviews process is over, but, using separate atomic commits during the process enable to see the history from the creation of the PR to the moment the review is done but I understand your point

internal/heartbeat/heartbeat.go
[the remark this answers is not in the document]
ydayagi: exactly. load and store. but it does not really matter

internal/heartbeat/heartbeat.go
ydayagi: One lock for entire update was not enough?

jakub-dzon (approved):

PR #157, opened 2022-05-24
Fix datarace when deregister and refactor

gabriel-farache:
This PR fixes the datarace when deregistering device detected byu  @jakub-dzon 

Basic refactoring was also performed:
- split methods into smaller methods
- reorganized/realigned structs fields of the registration package
- fix warnings

### DIFF
--- a/internal/registration/registration.go
+++ b/internal/registration/registration.go
@@ -38,16 +38,16 @@ type RegistrationWrapper interface {
 
 type Registration struct {
 	hardware         hardware2.Hardware
-	workloads        *workload.WorkloadManager
 	dispatcherClient pb.DispatcherClient
+	workloads        *workload.WorkloadManager
 	config           *configuration.Manager
-	registered       bool
-	RetryAfter       int64
-	deviceID         string
-	lock             sync.RWMutex
-	deregistrables   []Deregistrable
 	clientCert       *ClientCert
 	targetNamepsace  string
+	deviceID         string
+	deregistrables   []Deregistrable
+	RetryAfter       int64
+	lock             sync.RWMutex
+	registered       bool
 }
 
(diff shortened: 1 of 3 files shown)


internal/registration/registration_test.go
```diff
@@ -13,7 +13,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/big"
-	"os"
```
jakub-dzon, line 16: Why did you leave aliased `osUtil` instead of plain `os`?

internal/registration/registration.go
```diff
@@ -294,17 +311,16 @@ func (r *Registration) Deregister() error {
 		}
 	}
 
+	r.lock.RLock()
```
jakub-dzon, line 314: Is that the only fix for the race condition or are there other changes that are part of the fix?

internal/registration/registration_test.go
gabriel-farache: because `osUtil` alias was used and if someone put an alias there must a reason (I guess...) but yes, I agree with you that the alias is useless here, I can remove it if that good with you

internal/registration/registration.go
gabriel-farache: nope, that's the only fix

internal/registration/registration_test.go
jakub-dzon: Please remove it if `os` doesn't clash with any local variable

internal/registration/registration.go
jakub-dzon: :+1:

jakub-dzon (approved):

machacekondra (approved):

masayag (approved):

PR #160, opened 2022-05-24
Fix data race on startup

gabriel-farache:
This PR fixes the datarace on device startup detected byu @jakub-dzon

### DIFF
--- a/internal/logs/workload.go
+++ b/internal/logs/workload.go
@@ -2,6 +2,7 @@ package logs
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"git.sr.ht/~spc/go-log"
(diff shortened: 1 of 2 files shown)


machacekondra (approved):

internal/logs/workload.go
```diff
@@ -53,11 +54,12 @@ func (t *transportLog) Stop() {
...
-	transports      map[string]transportLog
-	wwriter         map[string]*WorkloadWriter // key is the workload name
-	workloadsConfig map[string]string
-	workloadManager *workload.WorkloadManager
+	deviceID            string
+	transports          map[string]transportLog
+	wwriter             map[string]*WorkloadWriter // key is the workload name
+	workloadsConfig     map[string]string
+	workloadManager     *workload.WorkloadManager
+	workloadsConfigLock sync.Mutex
```
masayag, line 62: Based on the use of this lock, will [RWMutex](https://pkg.go.dev/sync#RWMutex) be a better fit where it only locks for write in `Update` function, but allows multiple read in `WorkloadStarted` function?

gabriel-farache: I was not aware of such Mutex but generally speaking, yes, it would be better to use that kind of mutex here or even everywhere where there are more read locked than writes (see other PR with data races)

jakub-dzon: Please use the RWMutex instead.

masayag (approved):

internal/logs/workload.go
masayag: @jakub-dzon pls re-review

eloycoto (approved):

PR #163, opened 2022-05-26
Rootless podman

machacekondra:
This commit add support for rootless podman.

There are multiple things needed to support rootless podman:
 0) Create a user running the podman, flotta in our case.
 1) Enable linger for the user, so systemd can run even without session.
 2) Have set the /etc/subuid /etc/subgid for the user.
 3) Create home directory for the user, to store the data and systemd
    services.
 4) Enable podman.socket for the user.
 5) Volumes director must be owned by the user, as podman use it, not
    root.

Those prerequsites are done in spec file. For RPM based installation by
default the device-worker will execute podman as rootless. For dev-env
user will still use the rootfull pods. This could be changed the env
variables in the 'config/device-worker.toml'. We introduced two new env
variables:

  - HOME - store the home directory of the user
  - FLOTTA_XDG_RUNTIME_DIR - store the runtime dir of the user

FLOTTA_XDG_RUNTIME_DIR must be set for podman to properly work, this variable
is set by pam_systemd module, which is not used, as we don't switch the
user, we only use the podman socket of the user and the DBus connection,
which needs the env var as well.

The home directory is used for storing the podman data, and systemd
services.

The volume directory which is used by podman, must be owned by the user
which is manipulating the pods/containers, so that's done by Makefile
script.

Note the known limitation of rootless podman: https://github.com/containers/podman/blob/main/rootless.md

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,5 +23,5 @@ bin
 gover.coverprofile
 *.coverprofile
 
-# rpm
+# RPM
 *.tar.gz
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ gosec: ## Run gosec locally
 
 test: ## Run unit test on device worker
 test: test-tools
-	$(GINKGO)  --race -r $(GINKGO_OPTIONS) ./internal/* ./cmd/*
+	$(GINKGO) --race -r $(GINKGO_OPTIONS) ./internal/* ./cmd/*
 
 test-coverage:
 test-coverage: ## Run test and launch coverage tool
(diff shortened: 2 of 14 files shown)


machacekondra:
Please don't merge until I paste a link to successfull workflow job. Currently running here:
https://github.com/machacekondra/k4e-operator/runs/6624698003?check_suite_focus=true

machacekondra:
Verified here: https://github.com/machacekondra/k4e-operator/runs/6624698003?check_suite_focus=true

machacekondra:
Verified here: https://github.com/machacekondra/k4e-operator/runs/6624698003?check_suite_focus=true

masayag (approved):

jakub-dzon (approved):

PR #164, opened 2022-05-31
Logs function consumes 100% CPU

ydayagi:
transportLog in logs/workload.go sleeps for 100 milli when there's no more data
we need to do the same here
see test results here https://docs.google.com/spreadsheets/d/19AK-GxVl_egGC3bN1DekSf76p1CPMJoC05lmCRUJSAM/edit?usp=sharing

https://issues.redhat.com/browse/ECOPROJECT-821

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


jakub-dzon:
@ydayagi - what is the CPU load gain with that change? The stats for running with log collection in the linked results doc are missing.

ydayagi:
> @ydayagi - what is the CPU load gain with that change? The stats for running with log collection in the linked results doc are missing.

I updated the doc

pkliczewski (approved):

PR #165, opened 2022-06-06
do not close channels in function conatinerLog

ydayagi:
it is better to avoid closing channels

https://issues.redhat.com/browse/ECOPROJECT-837

### DIFF
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -515,8 +515,6 @@ func (p *podman) containerLog(ctx context.Context, containerID string, res io.Wr
 					log.Errorf("cannot write log line to io.Writer for container '%v': %v", containerID, err)
 				}
 			case <-ctx.Done():
-				close(stderrCh)
-				close(stdoutCh)
 				return
 			}
 		}


internal/workload/podman/podman.go
```diff
@@ -515,8 +515,6 @@ func (p *podman) containerLog(ctx context.Context, containerID string, res io.Wr
 					log.Errorf("cannot write log line to io.Writer for container '%v': %v", containerID, err)
 				}
 			case <-ctx.Done():
-				close(stderrCh)
```
tupyy, line 518: I think we should close the producer goroutine also. If you close only this one, data keeps accumulating into the channel for nothing.

ydayagi: data into the channel is not a problem. but if no one closes the routine then we may have a leak. I will check it.

ydayagi: @tupyy I opened a separate bug for this issue: https://issues.redhat.com/browse/ECOPROJECT-842

masayag (approved):

jakub-dzon (approved):

internal/workload/podman/podman.go
tupyy: @jakub-dzon yes

PR #168, opened 2022-06-07
Expose host devices os

tupyy:
This PR solves https://issues.redhat.com/browse/ECOPROJECT-832

It lists and parses all the host devices of `dev` folder without going into subfolders. 
The devices are added to immutable `hardware`.


tupyy:
pass it to WIP because I've found a better way to list devices

internal/hardware/hardware.go
```diff
@@ -108,3 +125,83 @@ func GetMutableHardwareInfoDelta(hardwareMutableInfoPrevious models.HardwareInfo
 
 	return hardwareInfo
 }
+
+// GetDevices recursively traverses a directory specified by path
+// and returns all devices found there.
+// Shamelessly copied from 	https://github.com/opencontainers/runc/blob/main/libcontainer/devices/device_unix.go
+func getDevices(path string) ([]*models.HostDevice, error) {
```
jakub-dzon, line 132: Have you considered contributing that code to [assisted installer agent](https://github.com/openshift/assisted-installer-agent)?

tupyy: yes. Initially, I used `Execute("ls"...)` and parse the result but it is not a solid solution. I think this one it is better

jakub-dzon: And why did you decide not to try to add that code to the assisted installer agent?

tupyy: you mean make a PR on installer agent repo?

jakub-dzon: Yes.

tupyy: I didn't think about it

internal/hardware/hardware_test.go
```diff
@@ -24,6 +24,7 @@ var _ = Describe("Hardware", func() {
```
pkliczewski, line 24: It would be great to add a test or two for something else than `Not(BeNil())`

tupyy: yes. but for that, we need to find a way to run tests as root.  Nevertheless, we need needs this for `mount` feature also.

gabriel-farache: Can't you mock the `getDevices` function? Because IMO, testing if the `getDevices` itself returns the correct list of devices on the host belong in the e2e tests rather than UT

tupyy: tu t'arrete pas!! cest la 2eme fois déjà :)

internal/hardware/hardware.go
```diff
@@ -108,3 +125,83 @@ func GetMutableHardwareInfoDelta(hardwareMutableInfoPrevious models.HardwareInfo
 
 	return hardwareInfo
 }
+
+// GetDevices recursively traverses a directory specified by path
+// and returns all devices found there.
+// Shamelessly copied from 	https://github.com/opencontainers/runc/blob/main/libcontainer/devices/device_unix.go
```
masayag, line 131: what's the penalty for adding it as a dependency and consume it directly? (`GetDevices` is an exported function)

tupyy: I've modified it a little bit to comply with our needs but I'm looking into it. maybe we can consume it as it is.

internal/hardware/hardware.go
```diff
@@ -108,3 +125,83 @@ func GetMutableHardwareInfoDelta(hardwareMutableInfoPrevious models.HardwareInfo
...
+	for _, f := range files {
+		switch {
+		case f.IsDir():
+			switch f.Name() {
+			// ".lxc" & ".lxd-mounts" added to address https://github.com/lxc/lxd/issues/2825
+			// ".udev" added to address https://github.com/opencontainers/runc/issues/2093
+			case "pts", "shm", "fd", "mqueue", ".lxc", ".lxd-mounts", ".udev":
+				continue
+			default:
+				sub, err := getDevices(filepath.Join(path, f.Name()))
```
masayag, line 147: is it safe from cyclic paths references? (an endless recursion that will lead to undesired result)

tupyy: no. I don't think so.

internal/hardware/hardware.go
tupyy: done

internal/hardware/hardware.go
```diff
@@ -111,7 +106,26 @@ func (s *HardwareInfo) GetMutableHardwareInfoDelta(hardwareMutableInfoPrevious m
...
+			Path:       d.Path,
+			DeviceType: string(d.Type),
+			Major:      d.Major,
+			Minor:      d.Minor,
+			UID:        int64(d.Uid),
+			Gid:        int64(d.Gid),
+		}
+
+		devices = append(devices, &device)
+	}
```
gabriel-farache, line 126: Could this be extracted into a private method `extractRuncDevices` (or something similar) or is that too much for you? (I tend to refactor as much as possible to have as small methods/functions as possible)

tupyy: I don't think it adds value. It's just a mapping.

jakub-dzon (approved):

PR #169, opened 2022-06-08
Bump _go_ version from 1.16.15 to 1.17.11 in github/workflows

tupyy:
Signed-off-by: Cosmin Tupangiu <cosmin@redhat.com>

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.15
+        go-version: 1.17.11
 
     - name: Install package
       run: |


eloycoto:
https://project-flotta.io/documentation/v0_1_0/dev/installation.html#device-installation

tupyy:
Thanks. That's putting me in trouble. I've bumped the version of the operator because of the commit which adds the models for the host devices but the new version bumps also the k8s dependencies version. :)
Any ideas :)?

eloycoto:
> Thanks. That's putting me in trouble. I've bumped the version of the operator because of the commit which adds the models for the host devices but the new version bumps also the k8s dependencies version. :)
> Any ideas :)?

So, for me two things: 

- Keep operator 1.16 compatible regarding apis/ and pkg, I think that was compatible. 
- Or move apis/ and pkg directories to another sub-project. 


I prefer the first one, but that it's a decision that needs to be taken at flotta-core meeting.

tupyy:
no more relevant. we keep 1.16

PR #171, opened 2022-06-10
Add disk sync usage metrics

jordigilh:
Adds prometheus metrics for data transfer:
* flotta_agent_datasync_files_transferred_counter{workload,direction}: Captures the number of files transmitted
* flotta_agent_datasync_bytes_transferred_counter{workload,direction}: Captures the number of bytes transmitted
* flotta_agent_datasync_time_transferred_counter{workload,direction}: Captures the time spent in completing the synchronization.
* flotta_agent_datasync_deleted_files_transferred_counter{workload,direction}: Captures the number of files deleted

This PR depends on https://github.com/project-flotta/flotta-operator/pull/259


cmd/device-worker/main.go
```diff
@@ -230,6 +231,9 @@ func main() {
 	go listenStartGracefulRebootChannel(wl, dataMonitor, systemMetricsWatcher, metricsStore, hbs, ansibleManager,
 		gracefulRebootChannel, deviceOs)
 
+	// Start prometheus httpd handler
+	go datatransfer.StartMetrics()
```
ydayagi, line 235: I do not see a reason to start a new goroutine and an HTTP server for collecting our own metrics.

internal/registration/gomock_reflect_616312687/prog.go
```diff
@@ -0,0 +1,66 @@
+
+package main
```
jakub-dzon, line 2: That file should not be commited.

internal/datatransfer/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+		}, []string{"workload_name", "direction"})
+	DeletedRemoteFilesTransferredCounter = factory.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "flotta_agent_datasync_deleted_files_transferred_counter",
+			Help: "The total number of files deleted from the target destination",
+		}, []string{"workload_name", "direction"})
+)
+
+func StartMetrics() {
+	http.Handle(DataSyncTransferMetricsPath, promhttp.Handler())
```
jakub-dzon, line 49: Why is HTTP server needed here?

internal/metrics/data_transfer.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+var defaultDataTransferMetricsConfiguration = models.ComponentMetricsConfiguration{Interval: DefaultDataTransferMetricsScrapingInterval}
+
+type DataTransferMetrics struct {
+	latestConfig atomic.Value
+	daemon       MetricsDaemon
+}
+
+func NewDataTransferMetrics(daemon MetricsDaemon) *DataTransferMetrics {
```
gabriel-farache, line 27: Where this is called? I can't find any references in the files changed on this PR

internal/metrics/data_transfer.go
```diff
@@ -0,0 +1,88 @@
...
+	if latestConfig != nil {
+		oldConfiguration, ok := latestConfig.(*models.ComponentMetricsConfiguration)
+		if !ok {
+			return fmt.Errorf("invalid type %+v", latestConfig)
+		}
+		if oldConfiguration != nil && reflect.DeepEqual(newConfiguration, *oldConfiguration) {
+			// Same configuration, no updates required.
+			return nil
+		}
+	}
```
gabriel-farache, line 53: Lines 43 to  53 could be extracted to a method `checkForChanges`

internal/datatransfer/monitor.go
```diff
@@ -135,13 +133,13 @@ func (m *Monitor) HasStorageDefined() bool {
...
 
-func (m *Monitor) SetStorage(fs FileSync) {
+func (m *Monitor) SetStorage(fs model.FileSync) {
 	m.syncMutex.Lock()
 	defer m.syncMutex.Unlock()
 	m.fsSync = fs
 }
 
-func (m *Monitor) getFsSync() (FileSync, error) {
+func (m *Monitor) getFsSync() (model.FileSync, error) {
```
gabriel-farache, line 142: This method needs to be refactored, at lead from lines 183 to 201 can be extracted (the for loop on the workloads)

gabriel-farache (commented):
Few refactors needed + I am wondering when metrics/data_transfer.go is used

cmd/device-worker/main.go
jordigilh: This routine is to expose the new metrics, not to collect them.

internal/datatransfer/metrics.go
jordigilh: Yes we could, but should we deviate from the standard pattern? what is the benefit here? I'm fine doing it if there is a significant gain, otherwise it just adds an exception to what one would expect.

internal/datatransfer/monitor.go
```diff
@@ -238,34 +231,64 @@ func containsDataPaths(dc *models.DataConfiguration) bool {
...
 		for _, dp := range dataConfig.Egress {
-			err := syncPath(syncWrapper, path.Join(hostPath, dp.Source), dp.Target)
+			err := m.syncPath(path.Join(hostPath, dp.Source), dp.Target)
 			if err != nil {
 				errors = multierror.Append(errors, err)
 			}
 		}
+		syncTime := time.Now().UnixMilli() - startSync
+		stats := syncWrapper.GetStatistics()
+		reportMetrics(workloadName, egress, stats.BytesTransmitted, stats.FilesTransmitted, stats.DeletedRemoteFiles, syncTime)
```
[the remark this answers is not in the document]
jordigilh: Will do.

internal/datatransfer/monitor.go
```diff
@@ -238,34 +231,64 @@ func containsDataPaths(dc *models.DataConfiguration) bool {
...
 		for _, dp := range dataConfig.Ingress {
-			err := syncPath(syncWrapper, dp.Source, path.Join(hostPath, dp.Target))
+			err := m.syncPath(dp.Source, path.Join(hostPath, dp.Target))
 			if err != nil {
 				errors = multierror.Append(errors, err)
 			}
 		}
+		syncTime := time.Now().UnixMilli() - startSync
+		stats := syncWrapper.GetStatistics()
+		reportMetrics(workloadName, ingress, stats.BytesTransmitted, stats.FilesTransmitted, stats.DeletedRemoteFiles, syncTime)
```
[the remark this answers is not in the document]
jordigilh: Yes, to the exception of the direction value: `ingress` or `egress`.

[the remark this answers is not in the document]
jordigilh: Will reuse the encapsulated function

internal/metrics/data_transfer.go
jordigilh: will check

internal/datatransfer/monitor.go
jordigilh: will check

cmd/device-worker/main.go
jordigilh: I see your point, but as I mentioned later on in @jakub-dzon comment, why should we deviate from the standard pattern? what is the benefit here? I'm fine doing it if there is a significant gain, otherwise it just adds an exception to what one would expect.

internal/registration/gomock_reflect_616312687/prog.go
jordigilh: Removed.

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
jordigilh: Refactored. Please take a look.

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
jordigilh: Refactored. Please take a look.

internal/metrics/data_transfer.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if newConfiguration.Disabled {
+		dt.daemon.DeleteTarget(dataTransferTargetName)
+	} else {
+		filter := getDataTransferSampleFilter(newConfiguration.AllowList)
+		dt.daemon.AddTarget(dataTransferTargetName, []string{datatransfer.MetricsEndpoint}, time.Duration(newConfiguration.Interval)*time.Second, filter)
+	}
```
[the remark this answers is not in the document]
jordigilh: Done

internal/datatransfer/monitor.go
jordigilh: Refactored a bit, what do you think now?

internal/metrics/data_transfer.go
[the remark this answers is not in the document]
gabriel-farache: Looks nice :)

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
gabriel-farache: Better but now I would extract the `for` loop (lines 259 to 265) into a method

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
gabriel-farache: Better but now I would extract the for loop (lines 259 to 265) into a method

internal/datatransfer/monitor.go
gabriel-farache: Better but why not extracting the `for` loop on the workload into another method (lines 184 to 194)?

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
jordigilh: If you don't have a strong opinion in this additional encapsulation, I'm fine the way it is. The function has a small footprint and it is readable as it is. I am not sure further encapsulation will help readability.

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
jordigilh: Same here, I'm fine the way it is now, I don't think further encapsulation will help in readability.

internal/datatransfer/monitor.go
gabriel-farache: Ok, good by me; I know I tend to create a lot a small methods every time I have something in an `if` or a `for`/`while` loop but we all have different way of coding :)

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
gabriel-farache: Ok, good by me; I know I tend to create a lot a small methods every time I have something in an `if` or a `for`/`while` loop but we all have different way of coding :)

internal/datatransfer/monitor.go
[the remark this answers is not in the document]
gabriel-farache: Ok, good by me; I know I tend to create a lot a small methods every time I have something in an `if` or a `for`/`while` loop but we all have different way of coding :)

internal/datatransfer/metrics.go
gabriel-farache: I think it would be good to quantify the amount of resources consumed by this handler, if it's idle and it does not consume anything, we should let it here, otherwise we would need to estimate the amount of resources consumed to create a metric and the frequency of the requests to get metrics.
In the device-worker, as we are in a resources' constrained environment, we need to be careful with them, I guess that why @jakub-dzon point that out

internal/datatransfer/monitor.go
jordigilh: Thanks for your understanding and for your feedback. In my opinion, the code looks much better after your suggestions than it did before it :smile:

internal/datatransfer/metrics.go
jordigilh: I'll check if we can call the Prometheus logic to serialize the metrics. Otherwise I'll quantify the extra memory of running the HTTP server inside a go program and report again. I don't think the HTTP server will consume much though but numbers talk for themselves :smile: 

If resource consumption are our concern, maybe we should look at splitting this part from the agent and running it as a container so we can limit the memory and CPU consumption and kill the process if it starts hogging too much memory. Knowing Prometheus a bit, it is the most common explanation why processes start consuming more memory over time, due to cardinality.

At the same time, if we're not doing it already, we should consider running node exporter as a containerized workload and align the metrics generated to those that are to be collected/white listed. This should reduce the memory used by node exporter compared to the default deployment, and as a bonus we also control how much memory it uses.
https://github.com/prometheus/node_exporter#collectors

Thoughts?

ydayagi: I do not understand the need to serialize metrics. Most applications create and collect their own metrics. Then these are exposed (exported) via a Prometheus client library. The library's internal implementation is an HTTP server that waits for Prometheus to scrape it. However, we do not have a Prometheus scraping us. We write to TSDB directly. So all you need is to periodically call our metrics 'write'/'add' method. If you generalize it a bit, then we could use it for all our features. Suppose you have a global struct with all metrics. Data upload, logs collection etc. Now suppose you have a timer that triggers every 60 seconds and writes the values from the struct to our TSDB.

cmd/device-worker/main.go
```diff
@@ -156,6 +157,14 @@ func main() {
 	}
 	configManager.RegisterObserver(systemMetricsWatcher)
 
+	dataTransferWatcher := metrics.NewDataTransferMetrics(metricsDaemon)
+	if err != nil {
```
masayag, line 161: there is no `err` returned by the call to `NewDataTransferMetrics`

internal/datatransfer/metrics.go
masayag: @ydayagi Is your suggestion to use https://github.com/project-flotta/flotta-device-worker/blob/be75e81fffbda7b0dcbc9c584535fdd0be214677/internal/metrics/daemon.go#L94 
in order to persist the collected sync-usage directly to TSDB and then use the existing remote write to push the metrics.
Any special consideration for this? does the write to TSDB need to be synced or verify it is initialized?

ydayagi: use Store, indeed. it is already synced by Prometheus's TSDB. And we already have concurrent calls to Store. Each target is a separate goroutine that calls Stor.

jordigilh: Removed httpd server and resolved to directly obtaining the metric data from the prometheus register. @ydayagi @jakub-dzon please take a look.

internal/datatransfer/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	DataSyncTransferMetricsAddress           = "localhost:9101"
+	DataSyncTransferMetricsPath              = "/metrics"
+	MetricsEndpoint                          = "http://" + DataSyncTransferMetricsAddress + DataSyncTransferMetricsPath
+	egress                         direction = "egress"
+	ingress                        direction = "ingress"
+)
+
+type direction string
```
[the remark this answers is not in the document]
jordigilh: No worries, I'll move it before the const.

cmd/device-worker/main.go
jordigilh: Done

jakub-dzon (changes requested):
Nice generalizing refactor :+1: 

Just a few more comments.

internal/metrics/daemon.go
```diff
@@ -120,16 +118,22 @@ func (tg *TargetMetric) Stop() {
...
-			log.Errorf("cannot start scraper for %v: %v", url, err)
+			log.Errorf("cannot start HTTP scraper for %v: %v", url, err)
 		}
 		scrapers = append(scrapers, scrape)
 	}
-	tg.scrapers = scrapers
+	return scrapers
+}
+
+func CreateObjectScraper(gather prometheus.Gatherer) Scrapers {
```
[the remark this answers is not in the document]
jordigilh: Moved.

jakub-dzon (approved):

Makefile
```diff
@@ -40,13 +40,11 @@ ifeq (, $(shell which ginkgo 2> /dev/null))
 	$(call go-install-tool,$(GOVER),github.com/sozorogami/gover)
 endif
 
-GINKGO = $(shell pwd)/bin/ginkgo
 ginkgo:
 ifeq (, $(shell which ginkgo 2> /dev/null))
```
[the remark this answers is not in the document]
jordigilh: You're right: if ginkgo is not installed this will fail. Do you know why do we install the binaries in `$(PROJECT_PATH)/bin` instead of `$(GOPATH)/bin`?
I'll revert the change, but it causes problems when I build it because I already have installed the binaries in `$(GOPATH)/bin` and the script can't find them there.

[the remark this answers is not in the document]
ydayagi: I don't know what was the original intention. I had issues with it as well.

internal/datatransfer/model/filesync.go
```diff
@@ -0,0 +1,15 @@
+package model
```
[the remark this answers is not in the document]
ydayagi: but it was all part of datatransfer package, no?

internal/metrics/scraper.go
```diff
@@ -93,3 +102,33 @@ func (s *HTTPScraper) Scrape(ctx context.Context) (model.Vector, error) {
...
+			for _, labels := range item.Label {
+				metric[model.LabelName(*labels.Name)] = model.LabelValue(*labels.Value)
+			}
+			v = append(v, &model.Sample{Metric: metric, Value: model.SampleValue(*item.Counter.Value), Timestamp: model.Now()})
+		}
+	}
+	return v, nil
+}
+
+func (o *ObjectScraper) String() string {
```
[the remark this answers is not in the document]
ydayagi: Not every type has or should have a string representation.  I think we should have called GetName or something like that. And what do u think about using a dynamic name? If I understand correctly, then we can have multiple object scrapers. Each with a different gatherer for a different feature. So perhaps there should be something unique for each instance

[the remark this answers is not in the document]
jordigilh: In this case it's not the name that we're showing, but the targets: so for HTTP scrapers we'll return the URL, but for the only object scraper we have we just return a static string. I considered returning the name of scraper or something more meaningful, but since we only have 1 single object scraper I didn't see the point.
But if we reach the day where we have more object scrapers we'll definitely need to return a more representative value, rather than the constant.

internal/datatransfer/model/filesync.go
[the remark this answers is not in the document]
jordigilh: Correct. It's because of import cycling issues between datatransfer and metrics:
```
package command-line-arguments
	imports github.com/project-flotta/flotta-device-worker/internal/metrics
	imports github.com/project-flotta/flotta-device-worker/internal/datatransfer
	imports github.com/project-flotta/flotta-device-worker/internal/datatransfer: import cycle not allowed

```

internal/metrics/scraper.go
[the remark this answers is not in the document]
jordigilh: Note that the `targetName` will return `data transfer` in this case, so there is not much else we can add to this debug entry.

internal/metrics/daemon.go
```diff
@@ -191,10 +190,10 @@ func (md *metricsDaemon) startTarget(target *TargetMetric) {
 }
 
 // AddTarget adds a metrics scraping target with filtering
-func (md *metricsDaemon) AddTarget(targetName string, urls []string, interval time.Duration, allowList SampleFilter) {
-	target := NewTargetMetric(targetName, interval, urls, md.store, allowList)
-	log.Debugf("added target '%v' with the following urls: '%+v", targetName, urls)
+func (md *metricsDaemon) AddTarget(targetName string, scrapers Scrapers, interval time.Duration, allowList SampleFilter) {
 
+	log.Debugf("added target '%v' with the following urls: '%s'", targetName, scrapers)
```
ydayagi, line 195: I think it should be scrapers instead of urls

internal/metrics/scraper.go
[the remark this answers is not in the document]
ydayagi: I wonder. we use scrapers and not a single scraper. do we have targets with multiple URLs?

ydayagi (approved):

internal/metrics/scraper.go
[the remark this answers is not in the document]
jordigilh: We do target multiple URLs for a workload, but each scrape will target a single workload, so the wording `urls` here is misleading. I have reworded it a bit, as well as the value returned from the `String()` function for the HTTP scraper to improve the entry.

internal/metrics/daemon.go
jordigilh: Updated. Please take a look.

internal/metrics/scraper.go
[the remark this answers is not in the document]
jordigilh: Updated.

PR #172, opened 2022-06-12
Fix issues with running device-worker

masayag:
The PR fixes several issues with device-worker:
* It adds executable permissions to .config folder:
  Without +x, no access can be made to the folder and files cannot be
  created in.
* Uses `id -u flotta` to obtain flotta user ID.
* Disables selinux as a workaround to failure to start podman.socket as
  flotta user. This issue is reported in
  bug https://bugzilla.redhat.com/show_bug.cgi?id=2060702
* Remove `systemctl user@...` since user systemd is controller by
  loginctl.

### DIFF
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -45,8 +45,12 @@ getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/no
 %post
 systemctl enable --now nftables.service
 loginctl enable-linger %{flotta_user}
-systemctl start user@$(getent passwd %{flotta_user} | cut -d: -f3).service
-systemctl --machine %{flotta_user}@.host --user start podman.socket
+
+# HACK till https://bugzilla.redhat.com/show_bug.cgi?id=2060702 is fixed
+setenforce 0
+systemctl enable --now --machine %{flotta_user}@.host --user podman.socket
+setenforce 1
+# END HACK
 
 %prep
 tar fx %{SOURCE0}
@@ -77,8 +81,13 @@ make install-worker-config USER=%{flotta_user} HOME=/home/%{flotta_user} LIBEXEC
 
 %post race
 loginctl enable-linger %{flotta_user}
-systemctl start user@$(getent passwd %{flotta_user} | cut -d: -f3).service
-systemctl --machine %{flotta_user}@.host --user start podman.socket
+
+# HACK till https://bugzilla.redhat.com/show_bug.cgi?id=2060702 is fixed
+setenforce 0
+systemctl enable --now --machine %{flotta_user}@.host --user podman.socket
+setenforce 1
+# END HACK
+
 systemctl enable --now nftables.service
 ln -sf %{_libexecdir}/yggdrasil/device-worker-race %{_libexecdir}/yggdrasil/device-worker
 
--- a/cmd/device-worker/main.go
+++ b/cmd/device-worker/main.go
@@ -41,7 +41,7 @@ const (
 func initSystemdDirectory() error {
 	systemddir := filepath.Join(os.Getenv("HOME"), ".config/systemd/user/")
 	// init the flotta user systemd units directory
-	err := os.MkdirAll(systemddir, 0644)
+	err := os.MkdirAll(systemddir, 0750)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


flotta-agent.spec
```diff
@@ -45,8 +45,12 @@ getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/no
 %post
 systemctl enable --now nftables.service
 loginctl enable-linger %{flotta_user}
-systemctl start user@$(getent passwd %{flotta_user} | cut -d: -f3).service
+systemctl start user@$(id -u %{flotta_user}).service
+
+# HACK till https://bugzilla.redhat.com/show_bug.cgi?id=2060702 is fixed
+setenforce 0
```
eloycoto, line 51: wait, here shouldn't we use enable instead of start?

masayag: @machacekondra can you refer to start vs enable?
do we have script for user flotta service?

flotta-agent.spec
```diff
@@ -45,8 +45,12 @@ getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/no
 %post
 systemctl enable --now nftables.service
 loginctl enable-linger %{flotta_user}
-systemctl start user@$(getent passwd %{flotta_user} | cut -d: -f3).service
+systemctl start user@$(id -u %{flotta_user}).service
```
machacekondra, line 48: This is nice, thanks!

flotta-agent.spec
machacekondra: Oh, right we should do both:
```
systemctl --machine %{flotta_user}@.host --user enable podman.socket
systemctl --machine %{flotta_user}@.host --user start podman.socket
```

flotta-agent.spec
```diff
@@ -45,8 +45,12 @@ getent passwd %{flotta_user} >/dev/null || useradd -g %{flotta_user} -s /sbin/no
...
+
+# HACK till https://bugzilla.redhat.com/show_bug.cgi?id=2060702 is fixed
+setenforce 0
 systemctl --machine %{flotta_user}@.host --user start podman.socket
```
machacekondra, line 52: O wonder if there is same selinux issue if using:
```
machinectl shell flotta@.host /usr/bin/systemctl start --user podman.socket
```

masayag: yes, same issue.

flotta-agent.spec
masayag: IIUC @eloycoto referred to the line above:
```
systemctl start user@$(id -u %{flotta_user}).service
```
vs
```
systemctl enable --now user@$(id -u %{flotta_user}).service
```

machacekondra: Yeah, `enable --now` is more elegant.

masayag: `systemctl start user@$(id -u %{flotta_user}).service` was removed.
it is being started by `loginctl enable`

pkliczewski: The bug was last updated on `2022-03-10`. It seems like it is not a priority to fix it. Maybe we could contribute a fix for this BZ?

pkliczewski (commented):
/lgtm

flotta-agent.spec
masayag: re-tested with 
```
machinectl shell %{flotta_user}@.host /usr/bin/systemctl enable --now --user podman.socket
```
it doesn't fail the port-script, however, it also doesn't enable the service.

flotta-agent.spec
```diff
@@ -77,8 +81,13 @@ make install-worker-config USER=%{flotta_user} HOME=/home/%{flotta_user} LIBEXEC
 
 %post race
 loginctl enable-linger %{flotta_user}
-systemctl start user@$(getent passwd %{flotta_user} | cut -d: -f3).service
```
machacekondra, line 80: Actually, we need to keep this. It starts the systemd for the user flotta, without it we can't start the podman.socket for the user.

machacekondra: So I just tested, and it seems, the podman.socket start, start the systemd service for the user, it seems the issues I had previously was because of the selinux. So it seems it's not needed anymore.

PR #173, opened 2022-06-14
[ECOPROJECT-745] "Port taken" error is not logged or propagated as an event

jordigilh:
Adds a port check before creating the systemd service.

### DIFF
--- a/internal/workload/wrapper.go
+++ b/internal/workload/wrapper.go
@@ -3,21 +3,23 @@ package workload
 import (
 	"context"
 	"fmt"
+	"net"
+	"os"
+	"syscall"
 
 	"io"
 	"sync"
 
-	"github.com/project-flotta/flotta-device-worker/internal/service"
-
 	"git.sr.ht/~spc/go-log"
+	_ "github.com/golang/mock/mockgen/model"
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/project-flotta/flotta-device-worker/internal/service"
 	"github.com/project-flotta/flotta-device-worker/internal/workload/api"
-	api2 "github.com/project-flotta/flotta-device-worker/internal/workload/api"
+
 	"github.com/project-flotta/flotta-device-worker/internal/workload/mapping"
 	"github.com/project-flotta/flotta-device-worker/internal/workload/network"
 	"github.com/project-flotta/flotta-device-worker/internal/workload/podman"
-	v1 "k8s.io/api/core/v1"
-
-	_ "github.com/golang/mock/mockgen/model"
 )
 
 const nfTableName string = "edge"
@@ -145,7 +147,7 @@ func (ww *Workload) Init() error {
 	return ww.netfilter.AddTable(nfTableName)
 }
 
-func (ww *Workload) List() ([]api2.WorkloadInfo, error) {
+func (ww *Workload) List() ([]api.WorkloadInfo, error) {
 	infos, err := ww.workloads.List()
 	if err != nil {
 		return nil, err
@@ -351,6 +353,9 @@ func getHostPorts(workload *v1.Pod) ([]int32, error) {
 	for _, c := range workload.Spec.Containers {
 		for _, p := range c.Ports {
 			if p.HostPort > 0 && p.HostPort < 65536 {
+				if !isPortAvailable(p.HostPort) {
+					return nil, fmt.Errorf("host port %d already in use for container %s in workload %s", p.HostPort, c.Name, workload.Name)
+				}
 				hostPorts = append(hostPorts, p.HostPort)
 			} else {
 				return nil, fmt.Errorf("illegal host port number %d for container %s in workload %s", p.HostPort, c.Name, workload.Name)
@@ -358,4 +363,28 @@ func getHostPorts(workload *v1.Pod) ([]int32, error) {
 		}
 	}
 	return hostPorts, nil
+}
+
+func isPortAvailable(port int32) bool {
+	conn, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
+	if err != nil {
+		errOpError, ok := err.(*net.OpError)
+		if !ok {
+			return true
+		}
+		errSyscallError, ok := errOpError.Err.(*os.SyscallError)
+		if !ok {
+			return true
+		}
+		errErrno, ok := errSyscallError.Err.(syscall.Errno)
+		if !ok {
+			return true
+		}
+		if errErrno == syscall.EADDRINUSE {
+			return false
+		}
+		return true
+	}
+	defer conn.Close()
+	return true
 }
--- a/internal/workload/wrapper.go
+++ b/internal/workload/wrapper.go
@@ -366,7 +366,7 @@ func getHostPorts(workload *v1.Pod) ([]int32, error) {
 }
 
 func isPortAvailable(port int32) bool {
-	conn, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
+	conn, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", port))
 	if err != nil {
 		errOpError, ok := err.(*net.OpError)
 		if !ok {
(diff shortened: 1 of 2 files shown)


internal/workload/wrapper.go
```diff
@@ -359,3 +364,27 @@ func getHostPorts(workload *v1.Pod) ([]int32, error) {
 	}
 	return hostPorts, nil
 }
+
+func isPortAvailable(port int32) bool {
+	conn, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
```
gabriel-farache, line 369: If it will always be on `localhost` that the port availability needs to be tested, maybe name the function `isLocalPortAvailable` for clarity of the intent and readability. If you feel that `Local` it is redundant with the context of the execution in a container, it's fine by me

machacekondra (approved):
PR looks great. The only problem I see is, that before we took those information from podman itself, now we need to create our own logic to test those things. It would be nice to have an issue which explore if we can get those information from podman itself.

internal/workload/wrapper.go
jakub-dzon, line 369: How will it work in our e2e tests, where the agent runs in a container?

machacekondra: We already do that, so should not be issue: https://github.com/project-flotta/flotta-operator/blob/066b8f6c2f527f01ff481b61b8fb8f519b0ff1be/test/e2e/e2e_test.go#L145

jakub-dzon (approved):

internal/workload/wrapper.go
jordigilh: Changed host address to be 0.0.0.0, that should make the context clear and also validate that the port is available across interfaces. What do you think now?

jordigilh:
> PR looks great. The only problem I see is, that before we took those information from podman itself, now we need to create our own logic to test those things. It would be nice to have an issue which explore if we can get those information from podman itself.

Agreed, that was my first intention but I realized that we can't do that when we use podman to [create the systemd service](https://github.com/project-flotta/flotta-device-worker/blob/66d54f0b29b3afec7e79d36528c8efcfb26338ae/internal/workload/wrapper.go#L240) to manage the pod's lifecycle. 

In that case we are left to check the journalctl logs or validate the systemd's service status and parse any errors that might happen, but still it would require to watch for changes in the service and then parse the log output. At that point we could do the same by watching for changes in the pod status in podman (react on degraded) and then try to figure it out in systemd's service status. A lot of extra layers that can go wrong anytime. 

If we could instead use the podman's API to start the container directly then we would be able to retrieve the errors coming from podman, sort of a canary testing.

If you have a better approach let me know, I'll be happy to implement it :smile:

internal/workload/wrapper.go
gabriel-farache: looks great! :)

gabriel-farache (approved):

internal/workload/wrapper.go
jordigilh: but I guess now the test will fail because the pod is not created since the port is taken, so no systemd service, right?

jordigilh:
> > PR looks great. The only problem I see is, that before we took those information from podman itself, now we need to create our own logic to test those things. It would be nice to have an issue which explore if we can get those information from podman itself.
> 
> 
> 
> Agreed, that was my first intention but I realized that we can't do that when we use podman to [create the systemd service](https://github.com/project-flotta/flotta-device-worker/blob/66d54f0b29b3afec7e79d36528c8efcfb26338ae/internal/workload/wrapper.go#L240) to manage the pod's lifecycle. 
> 
> 
> 
> In that case we are left to check the journalctl logs or validate the systemd's service status and parse any errors that might happen, but still it would require to watch for changes in the service and then parse the log output. At that point we could do the same by watching for changes in the pod status in podman (react on degraded) and then try to figure it out in systemd's service status. A lot of extra layers that can go wrong anytime. 
> 
> 
> 
> If we could instead use the podman's API to start the container directly then we would be able to retrieve the errors coming from podman, sort of a canary testing.
> 
> 
> 
> If you have a better approach let me know, I'll be happy to implement it :smile: 

On a second thought, i don't think this PR is going to do what we want it to do: the workload will remain as deploying and no error is going to be reported back. If we want to make it useful we should let the deployment go ahead and retrieve the status via podman (as @machacekondra was suggesting). Being port address taken is just another type of error (image not available, disk full, etc..) I don't believe we should deal with these kind of errors here, we should just report the phase status and, if possible, an error code or message back when possible, but always through the results of the systemd status. The reason is that there should be a single flow for reconciliation, any check we perform outside systemd and podman is another source of problems in the future.

machacekondra:
> On a second thought, i don't think this PR is going to do what we want it to do: the workload will remain as deploying and no error is going to be reported back. If we want to make it useful we should let the deployment go ahead and retrieve the status via podman (as @machacekondra was suggesting). Being port address taken is just another type of error (image not available, disk full, etc..) I don't believe we should deal with these kind of errors here, we should just report the phase status and, if possible, an error code or message back when possible, but always through the results of the systemd status. The reason is that there should be a single flow for reconciliation, any check we perform outside systemd and podman is another source of problems in the future.

We still run the podman [play kube](https://github.com/project-flotta/flotta-device-worker/blob/main/internal/workload/podman/podman.go#L296) command, because we need to generate the systemd services, from the pod. The output of this command should be also an error, for example:
```bash
$ podman play kube pod.yaml
[error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use]
[error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use error starting container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8: a dependency of container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8 failed to start: container state improper]
Pod:
374c9dc0abe7decd7d8634ca0211d39040accce02ed204f6d963e11e2fe74ff9
Container:
3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8

error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use
error starting container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8: a dependency of container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8 failed to start: container state improper

Error: failed to start 2 containers
```

I believe it's same for golang API. The reason it's reported is because the `--start` parameter is default for the pod. So we can catch error there - if I am not mistaken and I don't miss something.

jordigilh:
> > On a second thought, i don't think this PR is going to do what we want it to do: the workload will remain as deploying and no error is going to be reported back. If we want to make it useful we should let the deployment go ahead and retrieve the status via podman (as @machacekondra was suggesting). Being port address taken is just another type of error (image not available, disk full, etc..) I don't believe we should deal with these kind of errors here, we should just report the phase status and, if possible, an error code or message back when possible, but always through the results of the systemd status. The reason is that there should be a single flow for reconciliation, any check we perform outside systemd and podman is another source of problems in the future.
> 
> We still run the podman [play kube](https://github.com/project-flotta/flotta-device-worker/blob/main/internal/workload/podman/podman.go#L296) command, because we need to generate the systemd services, from the pod. The output of this command should be also an error, for example:
> 
> ```shell
> $ podman play kube pod.yaml
> [error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use]
> [error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use error starting container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8: a dependency of container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8 failed to start: container state improper]
> Pod:
> 374c9dc0abe7decd7d8634ca0211d39040accce02ed204f6d963e11e2fe74ff9
> Container:
> 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8
> 
> error starting container adc7d5260bf08d4b382f252ec79b127771a044368d886986f963227fb7ffe9db: rootlessport listen tcp 0.0.0.0:8886: bind: address already in use
> error starting container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8: a dependency of container 3fecdd0de736a465f32df6d51e06137758030de5c5cafde21d0df1186808d8a8 failed to start: container state improper
> 
> Error: failed to start 2 containers
> ```
> 
> I believe it's same for golang API. The reason it's reported is because the `--start` parameter is default for the pod. So we can catch error there - if I am not mistaken and I don't miss something.

I'll take a look from that angle, thanks :)

PR #175, opened 2022-06-16
fix panic/crash when deleting workload

ydayagi:
systemd struct has a field of type pointer to dbus conneciton
it is only assigned when creating a workload
however, it is used when deleting/updating a workload and upon startup
If you create a workload, restart the device-worker, update/delete a workload
then the device-worker will crash.

https://issues.redhat.com/browse/ECOPROJECT-859

### DIFF
--- a/internal/service/systemd.go
+++ b/internal/service/systemd.go
@@ -40,14 +40,14 @@ type Service interface {
 }
 
 type systemd struct {
-	Name           string            `json:"name"`
-	RestartSec     int               `json:"restartSec"`
-	Units          []string          `json:"units"`
-	UnitsContent   map[string]string `json:"-"`
-	servicePrefix  string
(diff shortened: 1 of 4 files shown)


internal/service/systemd.go
```diff
@@ -217,12 +210,20 @@ func (s *systemd) GetName() string {
 }
 
 func (s *systemd) reload() error {
-	return s.dbusConnection.ReloadContext(context.Background())
+	conn, err := newDbusConnection(s.Rootless)
```
machacekondra, line 213: I would run defer `conn.close()` as always open the connection

ydayagi: correct. forgot to add that. thanks

machacekondra (approved):

ydayagi:
@masayag @pkliczewski please review

jakub-dzon (approved):

PR #178, opened 2022-06-22
Add target to run rootfull test container

tupyy:
This PR adds a new target `test-docker` which executes all the test in a rootfull contiainer.

### DIFF
--- a/internal/workload/manager_test.go
+++ b/internal/workload/manager_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/project-flotta/flotta-device-worker/internal/common"
 	"github.com/project-flotta/flotta-device-worker/internal/workload"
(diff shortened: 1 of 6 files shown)


tools/Dockerfile_test
```diff
@@ -0,0 +1,10 @@
+FROM fedora:35
```
jordigilh, line 1: not f36?

tupyy: I wanted to have go1.16 but I can upgrade. No problem

internal/workload/manager_test.go
```diff
@@ -142,6 +143,10 @@ var _ = Describe("Manager", func() {
 			err = os.Chmod(datadir, 0444)
 			Expect(err).NotTo(HaveOccurred())
 
+			if common.IsContainerized() {
+				datadir = "/sys" // set to /sys which is always readonly
```
masayag, line 147: why creating a directory without write permissions will not work here in this case?

tupyy: because the containers runs with `CAP_DAC_OVERRIDE`

masayag (approved):

PR #179, opened 2022-06-22
RPM: Ostree changes 🚀

eloycoto:
When running on Ostree base operation system, wrting to /var is not
allowed at all, and things like enable linger for flotta agent is
not possible.

At the same time creating folders for user flotta was difficult because
when running it flotta user is not in place yet.

With this change, flotta-agent rpm is compatible with ostree OS, in this
case Red Hat enterprise Linux for Edge.

The main changes are:

- Adding sysusers.d file to enable users.
- Adding /etc/tmpfiles.d config to create files based on systemd

### DIFF
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -1,10 +1,10 @@
 %define _build_id_links none
-
+%global flotta_systemd packaging/systemd
 %global flotta_user flotta
 
 Name:       flotta-agent
 Version:    0.1.0
-Release:    1%{?dist}
+Release:    2%{?dist}
 Summary:    Agent application for the Flotta Edge Management solution
 ExclusiveArch: %{go_arches}
 Group:      Flotta
@@ -12,6 +12,7 @@ License:    ASL 2.0
 Source0:    %{name}-%{version}.tar.gz
 
 BuildRequires:  golang
+BuildRequires:  systemd-rpm-macros
 
 %if 0%{?rhel}
 Requires:       ansible-core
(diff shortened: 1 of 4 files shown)


flotta-agent.spec
```diff
@@ -73,28 +57,27 @@ install ./bin/device-worker %{buildroot}%{_libexecdir}/yggdrasil/device-worker
...
 %post race
-loginctl enable-linger %{flotta_user}
-
-# HACK till https://bugzilla.redhat.com/show_bug.cgi?id=2060702 is fixed
-setenforce 0
-systemctl enable --now --machine %{flotta_user}@.host --user podman.socket
-setenforce 1
-# END HACK
-
-systemctl enable --now nftables.service
```
machacekondra, line 95: What about the nftables? I can't see anywhere it to be enabled?

flotta-agent.spec
```diff
@@ -73,28 +57,27 @@ install ./bin/device-worker %{buildroot}%{_libexecdir}/yggdrasil/device-worker
 install ./bin/device-worker-race %{buildroot}%{_libexecdir}/yggdrasil/device-worker-race
 make install-worker-config USER=%{flotta_user} HOME=/home/%{flotta_user} LIBEXECDIR=%{_libexecdir} BUILDROOT=%{buildroot} SYSCONFDIR=%{_sysconfdir}
```
machacekondra, line 58: I think we need to change the `/home/%{flotta_user}` here to `/var/home/%{flotta_user}`

machacekondra:
Cool changes! Thanks! Can you please attach to commit msg or to PR link to documentation of this?

flotta-agent.spec
eloycoto: On rpm-ostree is doing anything at all, but I keep it for non ostree base os.

machacekondra (approved):

PR #180, opened 2022-06-23
Bump go 1.18

jordigilh:
Updated `go.mod` dependencies with `go get -u ./...`, run `go mod tidy` and bumped the go version to 1.18 in `go.mod`


eloycoto:
Related:

https://project-flotta.io/documentation/v0_1_0/dev/installation.html#device-installation

and 
https://github.com/project-flotta/flotta-device-worker/pull/169

Due to the project being now on low resources, I do not think that updating to 1.18 is something that we want to achieve.

PR #185, opened 2022-06-27
compare unordered workload lists

ydayagi:
the workload list does not gurantee order
to avoid considering two lists as not equal we have to compare unordered

https://issues.redhat.com/browse/ECOPROJECT-885


internal/configuration/configuration.go
```diff
@@ -202,3 +202,21 @@ func isEqualUnorderedSecretLists(x models.SecretList, y models.SecretList) bool
...
+	if len(x) != len(y) {
+		return false
+	}
+	workloadsMap := map[string]*models.Workload{}
+	for _, workload := range x {
+		workloadsMap[workload.Name] = workload
+	}
+	for _, workload := range y {
+		otherWorkload, ok := workloadsMap[workload.Name]
+		if !ok || !reflect.DeepEqual(workload, otherWorkload) {
```
eloycoto, line 217: but in this case, it'll fail on the podspec commands, no? I mean, there are plenty of arrays inside the json.

gabriel-farache: You mean the nested arrays inside fields of the `Workload` struct may now fail because of order issue, am I correct?
To list the problematic fields, it would be:

1. `Configmaps ConfigmapList` => contains array of `string`
2. `Data *DataConfiguration` => contains 2 arrayas 'Ingress and Egress) arrays of `DataPath` (which contains 2 `string` fields)

For `Labels map[string]string`, this one is a `map` and the order does (shall) not matter

ydayagi: there's nothing that will fail now and did not fail before. instead of deepEqual on the root (workloadsList), which does a deepEqual recursively, we do a deepEqual on each workload at a time

masayag: pls add unit-test. 
1. this method deserves it :-) 
2. it will certify the expected behavior for the complex objects in workload (e.g. arrays and/or maps)

gabriel-farache: > there's nothing that will fail now and did not fail before. instead of deepEqual on the root (workloadsList), which does a deepEqual recursively, we do a deepEqual on each workload at a time

Yes but the goal of your PR is to fix this behaviour, is it not? So it should correct all  failures related to the one your are fixing :)

eloycoto: I totally agree with @gabriel-farache that this PR should fix/check all the assumptions here, if not does not make sense to have the fix at all. IMHO.

masayag: perhaps it would be simpler (and gain same functionality) by calculating the hash of each structure, e.g. using https://github.com/mitchellh/hashstructure
and compare the result (we can store the existing hash of the existing configuration to save the need to recalculate it, aka premature optimization)

ydayagi: @masayag @eloycoto this fix is for a specific bug. The bug is that the workloads order changes and we do not take it into consideration. Let's focus on improving one step at a time.
By the way. the hash will not work. Why? same as DeepEqual. It is affected by order.

ydayagi: @gabriel-farache The behavior we are trying to fix is workloads order (this is the title of the PR). So this change does exactly that

gabriel-farache: > @gabriel-farache The behavior we are trying to fix is workloads order (this is the title of the PR). So this change does exactly that

@ydayagi you say in the description:

> the workload list does not gurantee order
> to avoid considering two lists as not equal we have to compare unordered

And to solve this, you compare each item of the list which each other to ensure the comparison of 2 equal `workload` lists will be correct. If that is not the goal then it's useless
But to compare correctly 2 `workload` items you can't use `deepEqual` as this struct has fields containing list field that may not guarantee the order

eloycoto: > @masayag @eloycoto this fix is for a specific bug. The bug is that the workloads order changes and we do not take it into consideration. Let's focus on improving one step at a time.
> By the way. the hash will not work. Why? same as DeepEqual. It is affected by order.
> 

This does not fix anything at all, and TBH, fix one case, but does not fix the issue.

ydayagi: what do u mean by 'anything at all'? did u encounter a situation where the two lists were considered not equal and the reason was not the order? So far we have not seen such behavior. We have seen and can reproduce, the workloads order problem. it is a serious bug. A bug that prevents adding logs for configuration changes.
So in reality, this fix adds an important capability that we are missing.

eloycoto: I gave reasons why this, in my opinion, is not completed, but you always hold the truth, so it's ok for me.

masayag (approved):

pkliczewski (approved):
I agree with @masayag. Let's have this in and later we can continue our improvements

gabriel-farache:
I feel it's weird to merge an incomplete work, especially when the span of the PR is not wide but whatever...

PR #190, opened 2022-06-29
CI: delete old copr builds

eloycoto:

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -36,6 +36,8 @@ jobs:
       - name: Submit the build by uploading the source RPM
         run: |
           make rpm-copr-testing
+          # Delete olders builds
+          copr-cli list-builds project-flotta/flotta-testing --output-format json | jq '[.[]|select(.name=="flotta-agent")] | sort_by(.id) | reverse[] | .id' | sed '1d' | xargs copr-cli delete-build
 
   container:
     runs-on: ubuntu-latest


masayag (approved):

.github/workflows/artifacts.yaml
```diff
@@ -36,6 +36,8 @@ jobs:
       - name: Submit the build by uploading the source RPM
         run: |
           make rpm-copr-testing
+          # Delete olders builds
+          copr-cli list-builds project-flotta/flotta-testing --output-format json | jq '[.[]|select(.name=="flotta-agent")] | sort_by(.id) | reverse[] | .id' | sed '1d' | xargs copr-cli delete-build
```
masayag, line 40: should we consider the status for the remained build? so at least one successful build is left (and delete the failed ones)

eloycoto: This is what `sed '1d'` is doing :-)

machacekondra (approved):

.github/workflows/artifacts.yaml
masayag: not sure, `sed '1d'` removes the first line. but couldn't it be that line is of a failed build?
is it guaranteed that only builds with status `succeeded` are in that list?

eloycoto: the last one is the one that was just created by the action, so if succeed executes that code :-)

PR #198, opened 2022-07-11
Fix agent installation

masayag:
The PR fixes errors of flotta-agent installation.
As part of the installation, flotta-agent service started and failed.
    
When selinux is enabled, post-script fails to create flotta user with error:
```
Jul 11 13:16:05 fedora useradd[3106]: new user: name=flotta, UID=1001, GID=1001, home=/var/home/flotta, shell=/sbin/nologin, from=none
Jul 11 13:16:05 fedora useradd[3106]: failed adding user 'flotta', exit code: 12
Jul 11 13:16:05 fedora bash[3106]: useradd: cannot create directory /var/home
```
   
The folder _/var/home/flotta_ requires updates to its selinux labels. To fix, we add
the label _user_home_dir_t_ to the folder  in order to serve as flotta's home directory.
    
In addition, certain flags changed for _useradd_ command to specify the base 
directory and also flag to instruct the creation of the home folder.

### DIFF
--- a/packaging/systemd/flotta-agent.service
+++ b/packaging/systemd/flotta-agent.service
@@ -7,8 +7,12 @@ Before=yggdrasild.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+# Since home directory is customed, we need to create it manually with specific selinux labels to allow access to it.
+ExecStart=bash -c "mkdir -p /var/home"
+ExecStart=bash -c "semanage fcontext -a -e /home /var/home"
+ExecStart=bash -c "restorecon -R -v /var/home"
 ExecStart=bash -c "getent group flotta >/dev/null || groupadd flotta"
-ExecStart=bash -c "getent passwd flotta >/dev/null || useradd -g flotta -s /sbin/nologin -d /var/home/flotta flotta"
+ExecStart=bash -c "getent passwd flotta >/dev/null || useradd -m -g flotta -s /sbin/nologin -b /var/home flotta"
 ExecStartPost=systemd-tmpfiles --create --remove --boot --exclude-prefix=/dev
 TimeoutSec=90s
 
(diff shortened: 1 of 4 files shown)


jordigilh:
does this impact rpm ostree based installations?

jordigilh (approved):

masayag:
> does this impact rpm ostree based installations?

Haven't (yet) verified. I don't expect it should. This will be verified :-)

packaging/systemd/flotta.conf
```diff
@@ -1,4 +1,4 @@
 F /var/lib/systemd/linger/flotta 0755 root root
+d /var/home/flotta 0777 flotta flotta -
```
machacekondra, line 2: If the order is important, can we document it?

packaging/systemd/flotta-agent.service
```diff
@@ -7,8 +7,12 @@ Before=yggdrasild.service
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+ExecStart=bash -c "mkdir -p /var/home"
+ExecStart=bash -c "semanage fcontext -a -e /home /var/home"
+ExecStart=bash -c "restorecon -R -v /var/home"
```
machacekondra, line 12: Can we document that those lines are just a workaround?

packaging/systemd/flotta.conf
masayag: it isn't important. it is to ease the reading (I usually preferred sorted files).
from the [documentation](https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html):

> When two lines are prefix path and suffix path of each other,
       then the prefix line is always created first, the suffix later
       (and if removal applies to the line, the order is reversed: the
       suffix is removed first, the prefix later)

packaging/systemd/flotta-agent.service
masayag: will do.

machacekondra (approved):

eloycoto (changes requested):

PR #199, opened 2022-07-12
Fix: greenboot command with rootless podman

eloycoto:
When doing an upgrade using Ostree the greenboot file is added
correctly, but with the rootless podman things started to fail because
it's not correctly defined by podman.

It's using sudo instead of machinectl because it's not part of RHEL 9
default.


masayag (approved):

bardielle (approved):
LGTM

PR #200, opened 2022-07-12
RPM: selinux added targeted context

eloycoto:
With this change selinux file will be added to be part of `/var/home/`
as home.

### DIFF
--- a/packaging/systemd/flotta-agent.service
+++ b/packaging/systemd/flotta-agent.service
@@ -9,8 +9,6 @@ Type=oneshot
 RemainAfterExit=yes
 # Since home directory is customed, we need to create it manually with specific selinux labels to allow access to it.
 ExecStart=bash -c "mkdir -p /var/home"
-ExecStart=bash -c "semanage fcontext -a -e /home /var/home"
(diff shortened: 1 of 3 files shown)


packaging/systemd/flotta-agent.service
```diff
@@ -9,8 +9,6 @@ Type=oneshot
 RemainAfterExit=yes
 # Since home directory is customed, we need to create it manually with specific selinux labels to allow access to it.
 ExecStart=bash -c "mkdir -p /var/home"
-ExecStart=bash -c "semanage fcontext -a -e /home /var/home"
```
masayag, line 12: this fails with error:
```
Error: Transaction test error:
  file /etc/selinux/targeted/contexts/files/file_contexts.subs from install of flotta-agent-0.1.0-3.fc35.x86_64 conflicts with file from package selinux-policy-targeted-35.3-1.20211019git94970fc.fc35.noarch
```
perhaps for ostree this can be mitigated by adding `|| true` not to fail for ostree?

masayag:
Closed in favor of https://github.com/project-flotta/flotta-device-worker/pull/202

PR #203, opened 2022-07-13
Use error interface instead of pointer

masayag:

### DIFF
--- a/internal/mount/mount.go
+++ b/internal/mount/mount.go
@@ -57,26 +57,26 @@ func (m *Manager) Update(config models.DeviceConfigurationMessage) error {
 	}
 
 	// holds the directories mounted in the current _Update_ call.
-	// It helps avoiding mounting the same directory twice.
+	// It helps to avoid mounting the same directory twice.
 	alreadyMounted := make(map[string]struct{})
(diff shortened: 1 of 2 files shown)


internal/mount/mount.go
```diff
@@ -57,26 +57,26 @@ func (m *Manager) Update(config models.DeviceConfigurationMessage) error {
...
 	// holds the directories mounted in the current _Update_ call.
-	// It helps avoiding mounting the same directory twice.
+	// It helps to avoid mounting the same directory twice.
 	alreadyMounted := make(map[string]struct{})
-	var merr *multierror.Error
+	var merr error
 	for _, mm := range config.Configuration.Mounts {
 		if _, found := alreadyMounted[mm.Directory]; found {
-			errf := fmt.Errorf("Directory '%s' has already been mounted. Skipping..", mm.Directory)
+			errf := fmt.Errorf("directory '%s' has already been mounted. Skipping directory", mm.Directory)
```
jordigilh, line 65: Since we can now mount devices in /dev, should it still be called directory? Perhaps mount path?

jordigilh (approved):

PR #206, opened 2022-07-14
Propagate annotations to podman's kube struct

jordigilh:
This change propagates the annotations defined in the workload object metadata that are prefixed `podman/` to podman's kube structure. The [changes](https://github.com/containers/podman/pull/13445) are aimed for podman v4.2.0. 

This change allows passing the crun annotation `run.oci.keep_original_groups:1` from the edgeworkload to podman and enable crun to keep the same host groupIDs  in the container. With this change, it is possible to access devices in `/dev/` using file system permissions.

There is a more modest workaround to this change, which is to add the annotation to all containers using the [configuration](https://github.com/containers/podman/blob/791458605af79e17e12a5a56fff30742ea0ff310/test/e2e/config/containers.conf#L56) in `containers.conf`.

Still, this change is not sufficient to allow interacting with certain devices as SELinux forbids it due to invalid access rules. 

For instance, in case of a webcam, SELinux still expects to have the following additional access rights:
```
module v4linux 1.0;

require {
	type v4l_device_t;
	type container_t;
	class chr_file { getattr ioctl map open read write };
}

#============= container_t ==============
allow container_t v4l_device_t:chr_file { getattr ioctl open read write };
allow container_t v4l_device_t:chr_file map;
```

There is a flag in SELinux that allows some operations by containers to devices (`container_use_devices`), but it lacks the `map` operation for the case of a webcam.

```bash
[root@localhost-live ~]# sesearch -A -s container_t -t v4l_device_t -c chr_file 
allow container_domain device_node:chr_file { append getattr ioctl lock open read write }; [ container_use_devices ]:True
``


jordigilh:
rpm build is failing because it won't build for f35 (golang 1.16). Can we move to f36 and deprecate f35?

jordigilh:
> No, Golang 1.16 is rhel 8 and is needed, I already revert commits for 1.17 twice, please keep 1.16 pray

Can we use rhel 9 then?

jordigilh:
@eloycoto @masayag I've removed the changes to migrate to go 1.17. Can you take another look?

masayag (approved):

masayag (commented):
I build rpm based on this PR and failed to verify it on fc36 with podman-4.2-rc1.
I had to add the devices's group-id to flotta user and disable selinux, which is the same things as using podman-4.1.

jordigilh:
> I build rpm based on this PR and failed to verify it on fc36 with podman-4.2-rc1.
> 
> I had to add the devices's group-id to flotta user and disable selinux, which is the same things as using podman-4.1.
> 
> 

Adding the supplementary group is needed in any scenario. The annotation instructs crun to propagate the same groupids from the flotta user to the container. The default is to create new group ids that don't match in id inside the container.

eloycoto (approved):

PR #208, opened 2022-07-20
Fix: Systemd add better logging

eloycoto:
For betting logging on the device level.


internal/service/systemd.go
```diff
@@ -187,11 +187,17 @@ func NewSystemdRootless(name string, units map[string]string, rootless bool) (Se
 }
 
 func (s *systemd) Add() error {
+	if len(s.UnitsContent) == 0 {
+		log.Infof("calling systemd add service with no units available")
```
masayag, line 191: is there an option to provide more context to this message, e.g. `s.Name`?

masayag (approved):

PR #212, opened 2022-07-21
Configure podman not to start the containers by default

jordigilh:
Fixes the problem where systemd refuses to start the service because the PID does not belong to root when running in a rootless container. The issue is described [here](https://github.com/containers/podman/issues/15032) for podman. The workaround is to create the containers but not start them by setting the flag `start` to false. 

Also, bumped spec release to 0.2.0-2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 VERSION = 0.2.0
-RELEASE = 2
+RELEASE = 1
 DIST_DIR = $(shell pwd)/dist
 CGO_ENABLED = 0
 OS :=$(shell awk -F= '/^ID/{print $$2}' /etc/os-release)
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 VERSION = 0.2.0
-RELEASE = 1
+RELEASE = 2
 DIST_DIR = $(shell pwd)/dist
 CGO_ENABLED = 0
 OS :=$(shell awk -F= '/^ID/{print $$2}' /etc/os-release)
--- a/internal/workload/podman/podman.go
+++ b/internal/workload/podman/podman.go
@@ -61,6 +61,7 @@ WantedBy=default.target
 var (
 	logTailLines = "1"
 	boolTrue     = true
+	boolFalse    = false
 )
 
 type AutoUpdateUnit struct {
@@ -318,6 +319,7 @@ func (p *podman) Run(manifestPath, authFilePath string, annotations map[string]s
 		Authfile:    &authFilePath,
 		Network:     &network,
 		Annotations: annotations,
+		Start:       &boolFalse,
 	}
 	report, err := play.Kube(p.podmanConnection, manifestPath, &options)
 	if err != nil {
--- a/flotta-agent.spec
+++ b/flotta-agent.spec
@@ -4,7 +4,7 @@
 
 Name:       flotta-agent
 Version:    0.2.0
-Release:    2%{?dist}
+Release:    1%{?dist}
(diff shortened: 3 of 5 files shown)


internal/workload/podman/podman.go
```diff
@@ -318,6 +319,7 @@ func (p *podman) Run(manifestPath, authFilePath string, annotations map[string]s
 		Authfile:    &authFilePath,
 		Network:     &network,
 		Annotations: annotations,
+		Start:       &boolFalse,
```
eloycoto, line 322: but if we do not start here, who is going to start it? systemd?

jordigilh: Correct.

eloycoto: Let's get @masayag to decide here, I'm worried a bit here, if `Enable()` or `Add()` fail, the container will not start at all, I'm happy to do it in this way, but maybe we added a corner case that I cannot see now.

jordigilh: If the workload does not run, the agent will attempt to run it again. The error will be reported:
https://github.com/project-flotta/flotta-device-worker/blob/278d2585a05d010b5353b6ac5482bf68c5cc3c0b/internal/workload/manager.go#L199-L212

Personally, I think it makes sense to be systemd to start the pod rather than podman itself. Systemd will take care of the retries. This leads to a similar scenario as you mention right now where the pod is running but either the `Enable()` or `Add()` triggers a failure, and is reported back as such. You will have a running workload (triggered by podman) and an error event stating that the workload failed due to either the `Enable()` or `Add()` operation. This is not consistent and this change actually fixes this scenario.

Makefile
```diff
@@ -1,5 +1,5 @@
 VERSION = 0.2.0
-RELEASE = 1
+RELEASE = 2
```
masayag, line 2: If I'm not wrong, this should be reserved only for PRs that release a new version

internal/workload/podman/podman.go
masayag: +1 for managing workloads by systemd.
Overrall, that's what we're actually doing with systemd and the reason we're using it.
In addition, this is the recommended method by podman to approach this case.
What we can do is making sure we monitor the workloads properly, but other than that - the pod management is delegated to systemd.

jordigilh: Following this topic, I discovered that by using systemd to start the containers we were not receiving the `pod started` events that are only triggered by podman when a pod starts by podman. Systemd triggers the `container started` event, which we don't capture. Because of that we were not triggering the workload observer loop to start scraping the metrics in the containers, for instance.
The quick workaround is to force a `podman.Start(podName)` after systemd has started it, so that we get the `pod started` event.
In principle, this does not impact the system as podman is not taking any action since the pod is already started (because all containers are running, I suppose), but it begs the question what would happen if one container is not running and systemd is restarting the service at the same time. Worst case this will happen only once as podman won't retry again.
For the time being I propose this as a temporal solution to the `pod started` event. For a better approach, I'd suggest to start capturing the `container started` event instead and introspect the pod status: If the pod status is running then we return the pod event `started`, otherwise we ignore the event. WDYT?
If you agree I will amend my commit with this approach, otherwise I will keep the `hack()` to trigger the `pod.Start()`.

Makefile
jordigilh: Understood. Will revert here and in the `.spec` file

jordigilh: Done

masayag (approved):

PR #215, opened 2022-07-26
Fix: Podman event when start

eloycoto:
Report events on the channel and avoid to run Run on podman to avoid
conflicts with Systemd.


internal/workload/podman/podman.go
```diff
@@ -281,7 +282,7 @@ func (p *podman) Events(events chan *PodmanEvent) {
 	}()
 }
 
-func (p *podman) getPodReportforId(podID string) (*PodReport, error) {
+func (p *podman) GetPodReportforId(podID string) (*PodReport, error) {
```
masayag, line 285: nit: if already changing it, `GetPodReportForId`

jordigilh (approved):
LGTM

masayag (approved):

PR #219, opened 2022-08-02
Fix: systemd workload disable

eloycoto:
When running on systemd some leftovers are on
`/var/home/flotta/.config/systemd/user/default.target.wants/` mainly
because service is deleted, but the symlink is not deleted at all.


masayag (approved):

masayag:
@eloycoto pls rebase

PR #223, opened 2022-08-04
WIP: Build: check that podman 4.2 is in place

eloycoto:

### DIFF
--- a/.github/workflows/rhel-rpm.yaml
+++ b/.github/workflows/rhel-rpm.yaml
@@ -24,4 +24,4 @@ jobs:
       - name: Build container fedora 36
         run: |
           docker build -t flotta/f36 -f tools/f36-Dockerfile .
-          docker run --rm -w /opt/device -v $(pwd):/opt/device flotta/f36 make rpm-build
+          # docker run -d --name flotta-f36 --rm -w /opt/device -v $(pwd):/opt/device flotta/f36 make rpm-build
--- a/tools/f36-Dockerfile
+++ b/tools/f36-Dockerfile
@@ -2,3 +2,16 @@ FROM fedora:36
 
 RUN dnf -y install rpm-build golang device-mapper-devel systemd-rpm-macros bash git make btrfs-progs-devel
 RUN git config --global --add safe.directory /opt/device
+
+COPY . /opt/device
+WORKDIR /opt/device
+
+RUN make rpm-build
+
+RUN dnf install -y 'dnf-command(copr)'
+RUN dnf copr enable project-flotta/flotta-testing -y
+
+RUN dnf install -y podman yggdrasil node_exporter ansible
+
+RUN rpm -i /root/rpmbuild/RPMS/x86_64/flotta-*.rpm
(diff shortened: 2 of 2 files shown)


eloycoto:
@jordigilh, what was the problem, it's using podman 4.2

jordigilh:
> @jordigilh, what was the problem, it's using podman 4.2

when I deploy the flotta agent it does not fail if the version of podman installed is not >= 4.2.0

.github/workflows/rhel-rpm.yaml
```diff
@@ -24,4 +24,4 @@ jobs:
       - name: Build container fedora 36
         run: |
           docker build -t flotta/f36 -f tools/f36-Dockerfile .
-          docker run --rm -w /opt/device -v $(pwd):/opt/device flotta/f36 make rpm-build
+          # docker run -d --name flotta-f36 --rm -w /opt/device -v $(pwd):/opt/device flotta/f36 make rpm-build
```
jordigilh, line 27: is this line necessary?

masayag:
I created https://github.com/project-flotta/flotta-device-worker/pull/226 to make sure the latest rpm versions are picked.
I can confirm that podman-4.2-rc1 is installed on the device.

jordigilh:
with `--best` it can find podman-4.2.0~rc1 from the copr repositories:
```
[root@fedora ~]# dnf  --best install podman --enablerepo=updates-testing
Fedora 36 - x86_64 - Test Updates                                                                                                                                                    2.5 MB/s | 9.9 MB     00:03    
Last metadata expiration check: 0:00:02 ago on Thu 04 Aug 2022 05:10:21 PM EDT.
Dependencies resolved.
=====================================================================================================================================================================================================================
 Package                                    Architecture                   Version                                        Repository                                                                            Size
=====================================================================================================================================================================================================================
Installing:
 podman                                     x86_64                         4:4.2.0~rc2-1.fc36                             copr:copr.fedorainfracloud.org:project-flotta:flotta-testing                          12 M
```

This is what is already available in the `install-agent-dnf.sh`

jordigilh:
The problem was with this line:
https://github.com/project-flotta/flotta-operator/blob/500935d6d8f2a6d4da046e97af9348c17c4325eb/hack/install-agent-dnf.sh.template#L57
But it's already commented out and it works fine when not installing podman prior to the flotta-agent. My understanding is that this line:
https://github.com/project-flotta/flotta-device-worker/blob/main/flotta-agent.spec#L29
should enforce the dependency and version, and in case not fulfilled it should either fail the installation or include the newer version of podman as part of it. Correct me if I'm wrong here:

* Install podman 4.1.1 (dnf install -y podman-4.1.1 to enforce it if necessary. I tested with a fresh f36 VM without copr project-flotta)
* Install flotta-agent

**Expected:**
The flotta-agent should install `podman-4.2.0~rc1` from the copr project-flota repository

**Actual:**
podman is not updated as part of the installation.

eloycoto:
Fixed by https://github.com/project-flotta/flotta-device-worker/pull/227

PR #231, opened 2022-08-08
Upgrade podman image version to v4.2.0-rc3

masayag:
Updated edgedevice image to use podman v4.2.0-rc3.
Build of base image done by using the following Dockerfile:
```
FROM quay.io/podman/upstream:latest
RUN dnf -y remove podman
RUN dnf -y install --best podman --enablerepo=updates-testing --disablerepo=copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next
```
Image pushed to quay.io/project-flotta/podman:v4.2.0-rc3-dev

Changes were made to edgedevice Dockerfile since the configuration file located in a different location and few tweaks were needed.

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -1,21 +1,26 @@
 # FROM quay.io/podman/stable:v4.2.0
 # Hack until 4.2 is released
-FROM quay.io/project-flotta/podman:v4.2.0-dev
+FROM quay.io/project-flotta/podman:v4.2.0-rc3-dev
 
 WORKDIR /project
 RUN dnf install -y 'dnf-command(copr)'
 RUN dnf copr enable project-flotta/flotta-testing -y
 
 # Yum dependencies
-RUN dnf update -y \
+RUN dnf update -y --exclude podman \
   && dnf install -y openssl procps-ng dmidecode nc iproute \
   yggdrasil flotta-agent-race node_exporter systemd-container \
   libselinux-utils util-linux sssd-client
 
 # Modify podman configuration
-RUN sed -i s/netns=\"host\"/netns=\"private\"/g /etc/containers/containers.conf && \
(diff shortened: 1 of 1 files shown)


tools/Dockerfile
```diff
@@ -1,21 +1,26 @@
...
+RUN dnf update -y --exclude podman \
   && dnf install -y openssl procps-ng dmidecode nc iproute \
   yggdrasil flotta-agent-race node_exporter systemd-container \
   libselinux-utils util-linux sssd-client
 
 # Modify podman configuration
-RUN sed -i s/netns=\"host\"/netns=\"private\"/g /etc/containers/containers.conf && \
-    sed -i s/utsns=\"host\"/utsns=\"private\"/g /etc/containers/containers.conf && \
-    sed -i s/ipcns=\"host\"/ipcns=\"private\"/g /etc/containers/containers.conf
+RUN sed -i 's/netns = \"host\"/netns = \"private\"/g' /usr/share/containers/containers.conf && \
```
gciavarrini, line 16: Why is it necessary to change from `/etc/containers/` to `/usr/share/containers`?
[Here](https://docs.podman.io/en/latest/markdown/podman.1.html#configuration-files) they say
>  "Distributions ship the /usr/share/containers/containers.conf file with their default settings. Administrators can override fields in this file by creating the /etc/containers/containers.conf file. Users can further modify defaults by creating the $HOME/.config/containers/containers.conf file. "

masayag: We build the edgedevice set on top of podman image runs on Fedora 36.
Then we modify it and create our own edgedevice image.

In the recent podman image, that's the location of the file. Hence, for our Dockerfile to work, we need to refer to the configuration file in its location.

gciavarrini: Thanks for the explanation!

gciavarrini (approved):
lgtm

PR #236, opened 2022-08-12
Use podman latest as base image for edgedevice

masayag:
The base image was updated in quay.io/project-flotta
which has the containers.conf under /etc/containers/container.conf

Also adding less utility to ease the debug.


masayag:
Apologizes for self-acking this PR.This fixes integration tests and dev-cli base image as well.

PR #240, opened 2022-09-03
return sync error

ahmadateya:


internal/datatransfer/s3/sync.go
```diff
@@ -107,5 +107,9 @@ func createHttpClient(caBundle []byte) *http.Client {
 
 func (s *Sync) SyncPath(sourcePath, targetPath string) error {
 	target := fmt.Sprintf("s3://%s/%s", s.bucketName, targetPath)
-	return s.s3SyncManager.Sync(sourcePath, target)
+	err := s.s3SyncManager.Sync(sourcePath, target)
+	if err != nil {
+		return fmt.Errorf("error while syncing path to s3: %v", err)
```
pkliczewski, line 112: There is already `error while` [here](https://github.com/project-flotta/flotta-device-worker/blob/main/internal/datatransfer/monitor.go#L285). Please remove it from here or there.

ahmadateya: ok, I removed my message, now I only return the error itself

internal/datatransfer/s3/sync.go
```diff
@@ -107,5 +107,9 @@ func createHttpClient(caBundle []byte) *http.Client {
 
 func (s *Sync) SyncPath(sourcePath, targetPath string) error {
 	target := fmt.Sprintf("s3://%s/%s", s.bucketName, targetPath)
-	return s.s3SyncManager.Sync(sourcePath, target)
+	err := s.s3SyncManager.Sync(sourcePath, target)
+	if err != nil {
+		return err
+	}
+	return nil
```
gciavarrini, line 114: I think this part isn't necessary.

the`return s.s3SyncManager.Sync(sourcePath, target)` already return the error or nil (if there is no error).

internal/datatransfer/monitor.go
```diff
@@ -282,8 +282,8 @@ func (m *Monitor) syncPath(source, target string) error {
 	log.Debug(logMessage)
 	err := m.fsSync.SyncPath(source, target)
 	if err != nil {
-		log.Errorf("error while %s", logMessage)
-		return fmt.Errorf("error while %s", logMessage)
+		log.Errorf("error while %s, %s", logMessage, err)
+		return fmt.Errorf("error while %s, %s", logMessage, err)
```
gciavarrini, line 286: To be consistent with other log messages,  can you please reword the log message as `error while %s; err: %s`

(See for example this [log message](https://github.com/project-flotta/flotta-device-worker/blob/6fb49625319fd4742cbb84f029ebca4c840bb90a/internal/datatransfer/monitor.go#L160))

gciavarrini (changes requested):

internal/datatransfer/s3/sync.go
ahmadateya: done; removed it

internal/datatransfer/monitor.go
ahmadateya: done

internal/datatransfer/s3/sync.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	"net/http"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/project-flotta/flotta-device-worker/internal/datatransfer/model"
 	"github.com/project-flotta/flotta-operator/models"
 	"github.com/seqsense/s3sync"
+	log "github.com/sirupsen/logrus"
```
pkliczewski, line 16: No need to update imports without any other changes.

pkliczewski (commented):
Looks good to me. @gciavarrini what do you think?

internal/datatransfer/s3/sync.go
gciavarrini, line 16: Last comment :)
There is no need to edit this file anymore.

gciavarrini (approved):
lgtm

pkliczewski (approved):

